PR #10, opened 2020-07-02
feat: include nimbus-build-system and initial version of the Makefile to generate a static library

richard-ramos:
I have not tested this in MacOS yet (I only have a linux machine available, so help with this is appreciated). 
Also, don't focus on the content/output of nim_status.nim yet (the files generated in build/). I'm creating a small test project to verify the output works.

Refer to README.md for the installation/usage instructions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,7 @@ nimcache
 .DS_Store
 TODO
 nim_status
+test/test
+data/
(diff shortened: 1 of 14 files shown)


siphiuel:
@richard-ramos Tried building it on macOS, it works OK.

PR #21, opened 2020-07-13
feat: add setup and teardown foreign thread  gc functions and nim-tests target

richard-ramos:


richard-ramos:
There weren't problems before but I read this here: https://nim-lang.github.io/Nim/backends.html#memory-management-thread-coordination: 

> When the NimMain() function is called Nim initializes the garbage collector to the current thread, which is usually the main thread of your application. If your C code later spawns a different thread and calls Nim code, the garbage collector will fail to work properly and you will crash.

I decided to add these functions to preventively mitigate this issue (althrough in the case of status-react it should be the same thread always, i guess?)

jrainville:
Ah ok. Better safe than sorry I guess

PR #42, opened 2020-08-17
feat: introduce Nim impl of generateAlias

### LINKED ISSUE #32: implement generateAlias in nim

michaelsbradleyjr:
Closes #32

---

This PR depends on #41 and on a commit in a [PR branch](https://github.com/status-im/status-go/pull/2015) in status-go, so it's labeled *don't merge yet* for the time being.

It's also important to note that because of this PR's introduction of `nim-secp256k1` as a dependency in `vendor/` it's necessary to switch from building/linking status-go as a static library to building/linking as a shared library.


michaelsbradleyjr:
I just realized I need some conditionals for Windows / Linux re: running the C tests; in the case of Windows need to augment `PATH` and in the case of Linux need to spec `LD_LIBRARY_PATH`. I'll do that shortly and force push.

PR #49, opened 2020-09-03
Expose StartWallet/StopWallet

richard-ramos:


michaelsbradleyjr (changes requested):
Also need changes in: https://github.com/status-im/nim-status/blob/expose-start/stop-wallet/src/nim_status/c/nim_status.nim

michaelsbradleyjr (approved):

PR #55, opened 2020-09-14
WIP: checking that startNode test run correctly in CI, don't merge

michaelsbradleyjr:
WIP: checking that startNode test run correctly in CI, don't merge


michaelsbradleyjr:
Tests passed and I manually checked the test output to check they ran properly with expected output.

PR #56, opened 2020-09-17
fix?: alloc error

richard-ramos:
See comments in https://github.com/status-im/nimPNG/pull/1


jrainville (approved):

michaelsbradleyjr:
Closing since we've known for awhile that the same memory corruption problem can happen independently of nimPNG.

PR #61, opened 2020-10-13
feat: integrate nim-sqlcipher

michaelsbradleyjr:
This PR integrates `nim-sqlcipher` with respect to the latest commit on that repo's `master` branch, which itself went through a long cycle of refactoring and dragged this effort along with it.

One change that had to be introduced today was removing the `-j${NPROC} NIMFLAGS="--parallelBuild:${NPROC}"` cli options from the GitHub Actions workflow. For reasons unknown, starting today (17 Nov 2020) those options resulted in build failures in GitHub Actions, specifically the building of upstream sqlcipher within this repo's `vendor/nim-sqlcipher/vendor/sqlcipher`. Oddly enough, those cli options do not cause any problems in the GitHub Actions workflow of the nim-sqlcipher repo itself.

The README for this repo needs to be updated with instructions and explanations (I'll create an issue for that); for now, take a look at the GitHub Actions workflow file to see how to pass e.g. `SSL_INCLUDE_DIR` and related variables to `make`.

PR #77, opened 2020-10-22
feat: settings

### LINKED ISSUE #62: implement getSettings with nim-sqlcipher
https://github.com/status-im/status-go/blob/4ecca1169d2e0a07b4e0cf5a7a3f4044216c678f/multiaccounts/accounts/database.go

richard-ramos:
fixes #62

### DIFF
--- a/nim_status/lib/settings.nim
+++ b/nim_status/lib/settings.nim
@@ -195,6 +195,8 @@ proc getSettings*(db: DbConn): Settings =
   webview_allow_permission_requests FROM settings WHERE synthetic_id = 'id'"""
 
   for r in rows(db, query):
+    echo $r
+
     result.userAddress = r[0].strVal.parseAddress
     result.chaosMode = toOption[bool](r[1])
     result.currency =  toOption[string](r[2])
--- a/nim_status/lib/settings.nim
+++ b/nim_status/lib/settings.nim
@@ -1,18 +1,19 @@
-import settings/[types, utils]
+import settings/types
+import settings/utils
 import web3/conversions
 import sqlcipher
 import json
 import options
 
-export Settings, SettingsEnum, `$`, `toSettings`
 
-proc newSettingsError(key, value: string): ref SettingsError =
-  (ref SettingsError)(msg: "Invalid setting: " & key & ": " & value)
+export types
 
-proc newSettingsError(key: SettingsEnum, value: string): ref SettingsError =
-  newSettingsError($SettingsEnum, value)
 
-proc createSettings*(db: DbConn, s: Settings, nodecfg: JsonNode) = # TODO: replace JsonNode by a proper NodeConfig object
+proc checkType(setting: SettingsEnum, value: auto, typeExpected: type) =
+  if not (value is typeExpected): 
+    raise (ref SettingsError)(msg: "Invalid setting: '" & $setting & "', type " & "'" & $typeExpected & "' expected")
+
+proc createSettings*(db: DbConn, s: Settings, nodecfg: JsonNode) = # TODO: replace JsonNode by a proper NodeConfig object?
   let query = """INSERT INTO settings (
                   address, currency, current_network, dapps_address,
                   eip1581_address, installation_id, key_uid,
@@ -45,20 +46,31 @@ proc createSettings*(db: DbConn, s: Settings, nodecfg: JsonNode) = # TODO: repla
             s.signingPhrase,
             (if s.walletRootAddress.isSome(): $s.walletRootAddress.get() else: ""))
   
-proc saveSetting*(db:DbConn, setting: SettingsEnum, value: string) =
-  discard
+proc saveSetting*(db:DbConn, setting: SettingsEnum, value: auto) =
+  case setting
+  of SettingsEnum.ChaosMode:
+    checkType(setting, value, bool)
+    db.exec("UPDATE settings SET chaos_mode = ? WHERE synthetic_id = 'id'", value)
+  of SettingsEnum.Currency:
+    checkType(setting, value, string)
+    db.exec("UPDATE settings SET currency = ? WHERE synthetic_id = 'id'", $value)
+  else: 
+    raise (ref SettingsError)(msg: "Setting: '" & $setting & "'cannot be updated")
 
-proc saveSetting*(db:DbConn, setting: string, value: string) =
-  discard
-  #[try: 
-    parseEnum(setting)
-  catch
-    newSettingsError(setting, value)]#
+  # TODO: add missing settings
+
+
+proc saveSetting*(db:DbConn, setting: string, value: auto) =
+  try: 
+    let s = parseEnum[SettingsEnum](setting)
+    saveSetting(db, s, value)
+  except:
+    raise (ref SettingsError)(msg: "Unknown setting: '" & $setting)
 
 proc getNodeConfig*(db: DbConn): JsonNode =
-  # TODO:
-  # "SELECT node_config FROM settings WHERE synthetic_id = 'id'"
-  discard
+  let query = "SELECT node_config FROM settings WHERE synthetic_id = 'id'"
+  for r in rows(db, query):
+    return r[0].strVal.parseJson
 
 proc getSettings*(db: DbConn): Settings =
   let query = """SELECT address, chaos_mode, currency, current_network, custom_bootnodes, custom_bootnodes_enabled, dapps_address, eip1581_address, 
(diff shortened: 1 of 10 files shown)


nim_status/lib/settings/utils.nim
```diff
@@ -0,0 +1,77 @@
...
+  if not self.hasKey(key) or self{key}.getStr == "":
+    result = none(Address)
+  else:
+    result = some(parseAddress(self[key].getStr))
+
+proc getOptionJsonNode*(self: JsonNode, key: string): Option[JsonNode] =
+  if not self.hasKey(key) or self[key].kind == JNull:
+    result = none(JsonNode)
+  else:
+    result = some(self[key])
```
emizzle, line 43: A generic alternative (to reduce boilerplate) might be:
```nim
proc getOption[T](self: JsonNode, key: string): Option[T] =
  if not self.hasKey(key) or self.to(T) == T():
    result = none(T)
  else:
    # handle special cases
    when T is Address:
      result = some(parseAddress(self[key].to(T))
    when T is int64
      result = some(self[key].getBiggestInt)
    # for all other (default) cases
    else:
      result = some(self[key].to(T))
```

nim_status/lib/settings/types.nim
```diff
@@ -0,0 +1,214 @@
...
+    # SendPushNotifications indicates whether we should send push notifications for other clients
+    sendPushNotifications* {.serializedFieldName("send-push-notifications?").}: Option[bool]
+    appearance* {.serializedFieldName("appearance").}: uint
+    useMailservers* {.serializedFieldName("use-mailservers?").}: bool
+    usernames* {.serializedFieldName("usernames").}: Option[JsonNode]
+    walletRootAddress* {.serializedFieldName("wallet-root-address").}: Option[Address]
+    walletSetUpPassed* {.serializedFieldName("wallet-set-up-passed?").}: Option[bool]
+    walletVisibleTokens* {.serializedFieldName("wallet/visible-tokens").}: Option[JsonNode]
+    wakuBloomFilterMode* {.serializedFieldName("waku-bloom-filter-mode").}: Option[bool]
+    webViewAllowPermissionRequests* {.serializedFieldName("webview-allow-permission-requests?").}: Option[bool]
```
emizzle, line 110: It is my understanding that `{.serializedFieldName.}` is only needed if `json-serialization` is used, unless there is something I'm missing.

emizzle: Another way to prevent string mistakes would be use the stringified enum values here, like so:
```nim
chaosMode* {.serializedFieldName($SettingsEnum.ChaosMode).}: Option[bool]
```

emizzle (commented):
Awesome work! Left a few comments on adding json-serialization haha.

emizzle (approved):
Very well done! 👏

nim_status/lib/settings.nim
```diff
@@ -0,0 +1,245 @@
...
+    result.useMailservers = r[38].intVal.bool
+    result.usernames = toOption[JsonNode](r[39])
+    result.appearance = r[40].intVal.uint
+    result.walletRootAddress = toOption[Address](r[41])
+    result.walletSetUpPassed = toOption[bool](r[42])
+    result.walletVisibleTokens = toOption[JsonNode](r[43])
+    result.wakuBloomFilterMode = toOption[bool](r[44])
+    result.wakuEnabled =toOption[bool](r[45])
+    result.webViewAllowPermissionRequests = toOption[bool](r[46])
+    break
```
emizzle, line 245: There is probably a very clever way of deserialising a `DbValue` in to a type that would prevent the need for this.

```nim
proc to*(dbVal: DbValue, val: var bool) =
  val = dbVal.boolVal

proc to*(dbVal: DbValue, val: var int) =
  val = dbVal.intVal

proc to*(dbVal: DbValue, val: var string) =
  val = dbVal.strVal

# add more overloads for other types

proc to[T]*(dbVal: DbValue, t: typedesc[T]): T
  if fieldPairs(default(T)).len == 0:
    dbVal.to(result) # should go to one of the above proc overloads
    return result
  result = T()
  let colIdx = 0
  for key, val in fieldPairs(result):
    val = dvVal[colIdx].to(val.type) # need to make sure var in fieldPairs is mutable (I don't think it is)

for r in rows(db, query):
  result = r.to(Settings)
  break
```

emizzle: Btw, would the `break` only set the first row value? Maybe that's the intention?

nim_status/lib/settings.nim
```diff
@@ -0,0 +1,245 @@
...
+  fleet, hide_home_tooltip, installation_id, key_uid, keycard_instance_uid,
+  keycard_paired_on, keycard_pairing, last_updated, latest_derived_path, log_level,
+  mnemonic, name, networks, notifications_enabled, push_notifications_server_enabled,
+  push_notifications_from_contacts_only, remote_push_notifications_enabled,
+  send_push_notifications, push_notifications_block_mentions, photo_path,
+  pinned_mailservers, preferred_name, preview_privacy, public_key, remember_syncing_choice,
+  signing_phrase, stickers_packs_installed, stickers_packs_pending, stickers_recent_stickers,
+  syncing_on_mobile_network, use_mailservers, usernames, appearance, wallet_root_address,
+  wallet_set_up_passed, wallet_visible_tokens, waku_bloom_filter_mode, waku_enabled,
+  webview_allow_permission_requests FROM settings WHERE synthetic_id = 'id'"""
```
jrainville, line 195: Wouldn't a `SELECT *` be simpler here?

nim_status/lib/settings/types.nim
```diff
@@ -0,0 +1,141 @@
...
+    SyncingOnMobileNetwork = "syncing-on-mobile-network?",
+    UseMailservers = "use-mailservers?",
+    Usernames = "usernames",
+    WalletRootAddress = "wallet-root-address",
+    WalletSetUpPassed = "wallet-set-up-passed?",
+    WalletVisibleTokens = "wallet/visible-tokens",
+    Appearance = "appearance",
+    WakuEnabled = "waku-enabled"
+    WakuBloomFilterMode = "waku-bloom-filter-mode",
+    WebviewAllowPermissionRequests = "webview-allow-permission-requests?"
```
jrainville, line 59: This is the same as the enum with have in `nim-status-client` right? I guess once this is merged, we could use this `enum` instead of the one in the client repo

nim_status/lib/settings.nim
richard-ramos: Yes, but then I do not know the order of the columns in order to populate the object attributes. (I could not find in sqlite docs if it's possible to refer to column names in the results of the query)

nim_status/lib/settings/types.nim
richard-ramos: Yeah!

nim_status/lib/settings.nim
richard-ramos: Yeah!

richard-ramos: > There is probably a very clever way of deserialising a `DbValue` in to a type that would prevent the need for this.

I'll try that approach, it looks neat!

michaelsbradleyjr:
This PR branch has been rebased against latest changes to this repo's [`feat/sqlcipher`](#61) PR branch

michaelsbradleyjr:
The test added by this PR is failing on Windows, but only when ssl, etc. are linked as shared libraries. I'm re-running the tests to check if it fails consistently in that scenario. I don't have a theory yet as to why it fails, but I also created PR #97 to check whether some bumps in `vendor/` that come with the waku v2 PR might sort it out.

For now I've added a *don't merge yet* label.

michaelsbradleyjr:
> but I also created PR #97 to check whether some bumps in `vendor/` that come with the waku v2 PR might sort it out

PR #97 fails in the same way:

```
[NimScript] exec: test/nim/build/settings
D:\a\nim-status\nim-status\test\nim\settings.nim(33) settings
D:\a\nim-status\nim-status\nim_status\lib\settings.nim(199) getSettings
D:\a\nim-status\nim-status\nim_status\lib\settings\utils.nim(14) toOption
D:\a\nim-status\nim-status\vendor\nimbus-build-system\vendor\Nim\lib\system\fatal.nim(49) sysFatal
Error: unhandled exception: 'intVal' is not accessible using discriminant 'kind' of type 'DbValue' [FieldError]
```

michaelsbradleyjr:
This PR's branch has been reset with the commits on #97's branch.

michaelsbradleyjr:
Rebased against `master` and force-pushed again.

PR #78, opened 2020-10-28
refactor: upgrade to waku v2, impl/test minimum viable usage

### LINKED ISSUE #59: start nim-waku v2 node in nim-status

### LINKED ISSUE #60: receive RAW messages with nim-waku v2 in nim-status

michaelsbradleyjr:
Implement `initNode` in `nim_status/lib/waku.nim` that can be used to setup a basic waku v2 node for use in a smoke test that demonstrates successful initialization, subscribing, and publishing. The implementation is intentionally barebones and is based on [`nim-waku/examples/v2/basic2.nim`](https://github.com/status-im/nim-waku/blob/master/examples/v2/basic2.nim).

As the implementation is fleshed out to be a proper [Status client](https://specs.status.im/spec/1), initialization will involve more pieces of the waku library. See for example [`nim-waku/examples/v2/chat2.nim`](https://github.com/status-im/nim-waku/blob/master/examples/v2/chat2.nim).

The modules `nim_status/lib/waku/config.nim` and `nim_status/lib/waku/node.nim` were deleted and the `saveAccountAndLogin` proc was removed from `nim_status/lib.nim`. The reason is that some of that code isn't needed for usage of waku v2 and it's too early to provide a shim for status-go's `saveAccountAndLogin`. It is not necessary that `nim_status/lib.nim, nim_status/lib/**` modules match up exactly to the code organization and calling conventions in status-go. Rather, it's only necessary that, as far as possible, our implementations can eventually be wrapped in shim procs that allow for a drop-in replacement, while on their own providing APIs that make for a good developer UX when used from other Nim modules.

Closes #59.
Possibly closes #60, waiting on feedback.


oskarth:
Nice to see this!

A general comment: I would leave nim-waku v1 in the code base and import v2 in addition to this. Back in July this was discussed, where it seemed like a good approach would be to have a "status-waku" thin layer/module in nim-status and then switch which one is used with feature flags. This would allow toggling between using Waku v1 and v2.

This would be useful for a few reasons:
- The API between v1 and v2 is different, and this module would be able to contain those differences
- It would allow clients, like mobile and desktop, to do a more seamless switch, e.g. going from using status-go waku v1 to nim-waku v1 (same API, Nim uncertainty), to nim-waku v2, all under flags.
- It would allow easy switching in desktop/status-nim between using Waku v1 and v2 as development matures for waku v2.

It is also worth noting that Waku v1 and Waku v2 can actually be run in parallel, which is needed in bridge nodes: https://github.com/status-im/nim-waku/pull/238/files by @kdeme

richard-ramos (approved):
Nice!

PR #79, opened 2020-10-29
feat: pass requests done via callRPC to infura

richard-ramos:


jrainville (approved):
Nice

michaelsbradleyjr (commented):
Nice work!

Questions:

From the perspective of someone using `callRPC` from another Nim module, would there be a "better" a approach than params/response using `JsonNode` or is it simply better to go with the flow of how `json_rpc/client` works?

In a similar vein, while eventually we may put callPRC shims in `nim_status/lib/shim.nim` that give the exact behavior (or as close as can be) of `status-go`, in `lib.nim, lib/**` we can enjoy more freedom to develop the library in a way that's really great when consuming it from another Nim module, e.g. instead of returning a `JsonNode` that contains an error code and message, we could raise an exception or use nim-results, that kind of thing. Thoughts?

Could the `string` method names be an enumeration (not sure if that's the right term), similar to e.g. the `LogLevel` levels in `nim-chronicles` (`DEBUG`, `INFO`, etc.)? It seems like that might help cut down on errors during development, tap better into autocomplete, etc.

richard-ramos:
> From the perspective of someone using `callRPC` from another Nim module, would there be a "better" approach than params/response being `JsonNode` or is it simply better to go with the flow of how `json_rpc/client` works?

when using it from another nim module, instead of dealing with JsonNode, it should be possible to do `web3_conn.provider.name_of_the_rpc_method(params)`. (i.e.: `web3.provider.eth_getTransactionCount(fromAddress, "latest")`). Something we must take into account is that some RPC methods should not be allowed. We have to figure out a way to do that.

> In a similar vein, while eventually we may put callPRC shims in `nim_status/lib/shim.nim` that give the exact behavior (or as close as can be) of `status-go`, in `lib.nim, lib/**` we can enjoy more freedom to develop the library in a way that's really great when consuming it from another Nim module, e.g. instead of returning a `JsonNode` that contains an error code and message, we could raise an exception or use nim-result, that kind of thing. Thoughts?

I agree. This implementation is kinda raw, and I'm sure that once we have a feature that requires usage of RPC methods, we'll settle in a more dev friendly way to obtain the results!

> Could the `string` method names be an enumeration (not sure if that's the right term), similar to e.g. the `LogLevel` levels in `nim-chronicles` (`DEBUG`, `INFO`, etc.)? It seems like that might help cut down on errors during development, tap better into autocomplete, etc.

Good idea!

richard-ramos:
Replaced by #114

PR #80, opened 2020-10-29
feat: open/close db connection and connect to web3 when calling a login()/logout() proc

richard-ramos:


nim_status/lib/accounts.nim
```diff
@@ -12,7 +12,7 @@ proc login*(accountData, password: string) =
   # TODO: determine where will the DB connection live. In the meantime I'm storing it into a global variable
   # TODO: determine where the web3 conn will live
 
-  let path = currentSourcePath.parentDir().parentDir().parentDir() & "/build/" & accountData
+  let path = currentSourcePath.parentDir() / accountData & ".db"
```
michaelsbradleyjr, line 15: I may have a misunderstanding but I think this change would put the database inside `nim_status/lib`. Instead of `currentSourcePath` can we use [`getCurrentDir`](https://nim-lang.org/docs/os.html#getCurrentDir)?

Also, is there a `&` and some `"` missing re: `/ accountData`?

richard-ramos: > I may have a misunderstanding but I think this change would put the database inside nim_status/lib. Instead of currentSourcePath can we use getCurrentDir?

I changed it to use `getCurrentDir` instead. In any case, this code will have to be removed from here once we implement `saveAccountAndLogin` which is the one supposed to do the DB initialization

> Also, is there a & and some " missing re: / accountData?

`/`  is an alias of `joinPath`.  `currentDir() / accountData & ".db" ` is equivalent to `joinPath(currentDir(), accountData & ".db")

michaelsbradleyjr: > `/` is an alias of `joinPath`. `currentDir() / accountData & ".db"` is equivalent to `joinPath(currentDir(), accountData & ".db")`

TIL. Very nice, thanks for the explanation!

PR #81, opened 2020-11-03
Feature/sr support

siphiuel:


.gitmodules
```diff
@@ -49,9 +49,6 @@
 [submodule "vendor/nim-eth"]
 	path = vendor/nim-eth
 	url = https://github.com/status-im/nim-eth.git
-[submodule "vendor/nim-secp256k1"]
```
richard-ramos, line 52: Why was nim-secp256k1 removed?

siphiuel: Because it's duplicated.

michaelsbradleyjr:
For the record, (1) I'm not sure how I closed this PR and (2) however it happened it was not done purposefully. It should be re-opened or a new one can be created to take its place. Sorry, wish I knew what happened!

siphiuel:
Re-created in #107

PR #82, opened 2020-11-05
feat: add createAccount function to create a random account

jrainville:
Adds the createAccount function to create a random account from  a random private Key.

Returns it as an Account object containing the private key, public key and address.

The function must be called with an `rng` object that should be generated on app start so that it can be reused  avoid using system resources (such as urandom) after that (explanation taken from the nim-eth repo

### DIFF
--- a/nim_status.nimble
+++ b/nim_status.nimble
@@ -78,3 +78,4 @@ task tests, "Run all tests":
   buildAndRunTest "mailservers"
   buildAndRunTest "bip32"
   buildAndRunTest "contacts"
+  buildAndRunTest "account"
--- a/nim_status/lib.nim
+++ b/nim_status/lib.nim
@@ -1,5 +1,6 @@
 import lib/alias
 import lib/alias/data
+import lib/account
 import lib/identicon
 import lib/util
 import nimcrypto
@@ -7,6 +8,8 @@ import strformat
 import strutils
 import unicode
 import json
+
+export createAccount 
 
 proc hashMessage*(message: string): string =
   ## hashMessage calculates the hash of a message to be safely signed by the keycard.
--- a/nim_status/lib/account.nim
+++ b/nim_status/lib/account.nim
@@ -2,7 +2,7 @@ import
   std/[parseutils, strutils],
   normalize,
   secp256k1,
-  stew/[results],
+  stew/[results, byteutils],
   nimcrypto/[sha2, pbkdf2, hash, hmac],
   account/[types, paths],
   eth/keys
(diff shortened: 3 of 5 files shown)


src/nim_status.nim
```diff
@@ -4,7 +4,8 @@ from nim_status/types import
 from nim_status/lib/shim as nim_shim import
   hashMessage,
   generateAlias,
-  identicon
+  identicon,
+  createAccount
```
michaelsbradleyjr, line 8: At this time, new pure Nim implementations should probably not be exposed at the top-level *unless* they're accompanied by tests that show the Nim replacement gives identical output for the same input compared to the Go implementation. Instead they can be imported (for usage, testing) from a nim module in `lib/`. No big change needed, just delete this. Also, if we did keep this we'd need to delete the `createAccount` imported/exported from the go shim.

Also, there has been some reorg and it was encouraged on the call earlier this week that the `feat/settings` branch be used as the base. I guess we can live with it being based off `master` since it's not a big PR.

michaelsbradleyjr (changes requested):
Nice work! Just a few points below.

src/nim_status.nim
michaelsbradleyjr: As an example of "gives identical output..." tests see what's done for `generateAlias`:

https://github.com/status-im/nim-status/blob/master/tests/c/shims.c#L84

https://github.com/status-im/nim-status/blob/master/tests/nim/shims.nim#L84

michaelsbradleyjr: One thing re: "gives identical output"... for this function that's a challenge unless both the Go and Nim implementations allow a seed to optionally be supplied to make the generation deterministic. That may not be possible — I'd need to look at the status-go API and what's possible with the Nim libs we're using. If it's not possible then, that's fine, but we'd want to come up with a test that demonstrates that Nim and Go implementations have the same shape or something like that.

jrainville:
@michaelsbradleyjr updated and rebased on settings. Let me know if it's better

nim_status/lib/shim.nim
```diff
@@ -8,3 +8,5 @@ proc hashMessage*(message: string): string =
 export generateAlias
 
 export identicon
+
+export createAccount
```
michaelsbradleyjr, line 12: I didn't notice this previously or it's new. So, the thing is that procs exported in the shim are supposed to be 100% 1-to-1 with the input/output of the same function in status-go. It's perfectly fine if that's not the case for the procs in lib, but the shim proc would need to account for that.  If we don't want to take time now to implement the shim proc, that's okay, but the lib proc shouldn't simply be re-exported as is, so for now I think it should be removed. Does that make sense?  If not, please let me know because I may be doing a bad job of explaining the library's overall design.

michaelsbradleyjr (changes requested):

jrainville:
@michaelsbradleyjr I just rebased and your comments should be addressed now

michaelsbradleyjr (approved):

PR #84, opened 2020-11-05
custom tokens CRUD

iurimatias:


michaelsbradleyjr (changes requested):
Looks good! Just one thing: there's a spurious `test/test` binary included in the commit. I'm not sure where that came from but I don't think it's a matter of adding an entry to `.gitignore`; instead it seems like a `make` command misbehaved, not sure yet.

PR #86, opened 2020-11-12
feat: bip32

richard-ramos:
WIP

@stefantalpalaru:
In this project we have `vendor/nim-secp256k1` and we also have `vendor/nim-waku/vendor/nim-secp256k1`. When doing a `make test-nim`, nimbus-build-system ignores the folder `vendor/nim-secp256k1` and uses the submodule inside `nim-waku/vendor`. Is there a way to force nimbus-build-system to use the top level submodule? (I made a change in https://github.com/status-im/nim-secp256k1/pull/25 that is not being found)

### DIFF
--- a/nim_status/lib/account.nim
+++ b/nim_status/lib/account.nim
@@ -1,0 +1,62 @@
+import
+  std/[parseutils, strutils],
+  normalize,
+  secp256k1,
+  stew/[results],
+  nimcrypto/[sha2, pbkdf2, hash, hmac],
+  account/[types, paths]
+
+export KeySeed, Mnemonic, SecretKeyResult, KeyPath
+
+proc getSeed*(mnemonic: Mnemonic, password: KeystorePass = ""): KeySeed =
+  let salt = toNFKD("mnemonic" & password)
+  KeySeed sha512.pbkdf2(mnemonic.string, salt, 2048, 64)
+
+proc child(self: ExtendedPrivKey, child: PathLevel): ExtendedPrivKeyResult =
+  var hctx: HMAC[sha512]
+  hctx.init(self.chainCode)
+  if child.isNonHardened():
+    hctx.update(self.secretKey.toPublicKey().toRawCompressed())
+  else:
+    hctx.update([0.byte])
+    hctx.update(self.secretKey.toRaw())       
+  hctx.update(child.toBEBytes());
+  let hmacResult = hctx.finish();
+   
+  var secretKey = hmacResult.data[0..31]
+  let chainCode = hmacResult.data[32..63]
+
+  var pk = self.secretKey.toRaw()[0..^1]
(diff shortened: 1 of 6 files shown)


richard-ramos:
cc: @zah and @arnetheduck for ideas on how to improve this functionality :)

stefantalpalaru:
You could change the content of "vendor/.nimble/pkgs/nim-secp256k1-#head/nim-secp256k1.nimble-link" in the "deps" target of your Makefile: https://github.com/status-im/nim-status/blob/d1f4fa39d786b4266481999f4bb20efa88df2158/Makefile#L131

Look at this for inspiration: https://github.com/status-im/nimbus-build-system/blob/92e5042667b747d22106f085eaa9b5e9766ba474/makefiles/targets.mk#L117

Makefile
```diff
@@ -90,6 +90,10 @@ nat-libs-sub:
 		$(ENV_SCRIPT) $(MAKE) USE_SYSTEM_NIM=1 nat-libs
 
 deps: | deps-common nat-libs nat-libs-sub
+		echo "Updating nim-secp256k1 path"; \
+		SECP256K1_DIR=$$(readlink -f ./vendor/nim-secp256k1/); \
+		echo $$SECP256K1_DIR > vendor/.nimble/pkgs/nim-secp256k1-#head/nim-secp256k1.nimble-link; \
```
zah, line 95: Why is this necessary? We don't have such a step in the other projects using nim-secp256k1

nim_status/lib/account.nim
```diff
@@ -0,0 +1,62 @@
+import
```
zah, line 1: Wouldn't it make more sense for this code to be added to the `nim-eth` library or even a stand-alone nim-bip32 library?

Makefile
richard-ramos: Without this, when I import secp256k1 it will use `./vendor/nim-waku/vendor/nim-secp256k1` instead of `./vendor/nim-secp256k1`

michaelsbradleyjr: I think we're going to have to reconcile this some other way. It seems to cause a problem when compiling on Windows:
```
D:\a\nim-status\nim-status\nim_status\lib\alias.nim(2, 8) Error: cannot open file: secp256k1
```

Can we reconcile the version of nim-secp256k1 used by us at the top-level and the one in waku's vendor?

michaelsbradleyjr: With the changes in #115 we shouldn't need to do this.

michaelsbradleyjr (changes requested):

michaelsbradleyjr (approved):

michaelsbradleyjr:
I'm having GHA rerun the builds/tests one more time just to make sure they consistently pass.

PR #125, opened 2020-12-21
fix: use settings dbColumnName and dbTableName values

emizzle:
Use the Settings type's dbColumnName and dbTableName values in the `createSettings` query instead of hardcoding the names. This allows for changes, is strongly-typed, and prevents errors.

### DIFF
--- a/nim_status/lib/settings.nim
+++ b/nim_status/lib/settings.nim
@@ -10,15 +10,30 @@ import # nim-status libs
 export types, options
 
 proc createSettings*(db: DbConn, s: Settings, nodecfg: JsonNode) = # TODO: replace JsonNode by a proper NodeConfig object?
-  let query = """INSERT INTO settings (
-                  address, currency, current_network, dapps_address,
-                  eip1581_address, installation_id, key_uid,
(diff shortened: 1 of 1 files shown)


nim_status/lib/settings.nim
```diff
@@ -10,15 +10,30 @@ import # nim-status libs
...
-  let query = """INSERT INTO settings (
-                  address, currency, current_network, dapps_address,
-                  eip1581_address, installation_id, key_uid,
-                  keycard_instance_uid, keycard_paired_on,
-                  keycard_pairing, latest_derived_path, mnemonic,
-                  name, networks, node_config, photo_path,
-                  preview_privacy, public_key, signing_phrase,
-                  wallet_root_address, synthetic_id)
-                VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,'id')"""
+  var setting: Settings
```
jrainville, line 13: Just for my own knowledge, how does this work? You don't seem to initialize the variable, so how come below you can use its attributes?

PR #129, opened 2021-01-18
Transaction history

siphiuel:


michaelsbradleyjr:
Should this PR stay open and be rebased? Or with the new nim-status roadmap does it make sense to close it for now and revisit later?

cc: @iurimatias

Issue #131, opened 2021-01-21
disable shim and make initial status-desktop branch that uses nim-status exclusively


iurimatias:
this task is required for the subsequent upcoming tasks

goals/tasks:

1. nim-status: disable the shim to status-go (this will break status-desktop temporarily)
2. status-desktop: disable breaking calls until login screen works again

PR #145, opened 2021-01-26
Disable go-shim

richard-ramos:
Partially fixes: #131

### DIFF
--- a/nim_status.nim
+++ b/nim_status.nim
@@ -6,7 +6,7 @@ from nim_status/lib/shim as nim_shim import
   generateAlias,
   identicon
 
-from nim_status/dummy/shim as dummy_shim import
+from nim_status/go/shim as go_shim import
   initKeystore,
   openAccounts,
   multiAccountGenerateAndDeriveAddresses,
@@ -14,14 +14,16 @@ from nim_status/dummy/shim as dummy_shim import
   multiAccountImportMnemonic,
   multiAccountImportPrivateKey,
   multiAccountDeriveAddresses,
-  saveAccountAndLogin,
+  saveAccountAndLogin
+
+from nim_status/dummy/shim as dummy_shim import
+  login,
   deleteMultiAccount,
   callRPC,
   callPrivateRPC,
   addPeer,
   setSignalEventCallback,
   sendTransaction,
-  login,
   logout,
   verifyAccountPassword,
   validateMnemonic,
--- a/nim_status.nim
+++ b/nim_status.nim
@@ -6,7 +6,7 @@ from nim_status/lib/shim as nim_shim import
   generateAlias,
   identicon
 
-from nim_status/go/shim as go_shim import
+from nim_status/dummy/shim as dummy_shim import
   initKeystore,
   openAccounts,
   multiAccountGenerateAndDeriveAddresses,
--- a/nim_status/dummy/shim.nim
+++ b/nim_status/dummy/shim.nim
@@ -1,0 +1,171 @@
+import ../c/go/shim as go_shim
+import ../types
+
+export SignalCallback
+
+# All procs start with lowercase because the compiler will also need to import
+# status-go, and it will complain of duplication of function names
+
+proc throwEx() =
+  writeStackTrace()
+  raise newException(Defect, "DISABLED")
+
+proc hashMessage*(message: string): string =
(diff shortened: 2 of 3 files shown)


jrainville (approved):

michaelsbradleyjr (commented):
I think this is going more or less the right direction, but I would frame it a bit differently.

I'd like to have a discussion before proceeding with these changes. After this write-up, I'm going to work on a separate PR that demonstrates everything proposed below.

---

I think it should work like this:

**(1)** If you're consuming `nim-status` with Nim and *right now* you want the APIs and behavior to be *guaranteed* 1-1 with status-go, then you should `import nim_status/go/shim`.

Explanation: that shim is only a thin wrapper around status-go, so effectively using that shim is the same thing as if in the consuming Nim code you used `importc` pragma and linked to status-go yourself.

**(2)** If you're consuming `nim-status` as a C library (e.g. status-react on Android, iOS) and *right now* you want the APIs and behavior to be *guaranteed* 1-1 with status-go, then you should compile `nim_status/c/go/shim.nim` into `nim_status.a` and link to it.

Explanation: same as for (1), but noting that `nim_status/c/go/shim.nim` makes use of e.g. `cstring` instead of Nim's `string` to avoid problems with garbage collection.

**(3)** If you're consuming `nim-status` with Nim and want all the *functionality* of status-go (and more!) but don't want/need API compatibility with status-go, then you should `import nim_status`.

Explanation: top-level `nim_status.nim` re-exports all of the implementations in `nim_status/lib/*.nim`. Those APIs embrace Nim best-practices, are built around nim-chronos `async` pragma, nim-task-runner, etc. They do everything status-go does (and more!) but the names of the functions, how you call them, etc. are an independent implementation and don't necessarily match up with how status-go is used and organized. That's on purpose because `nim_status/lib` is the future and the APIs and implementations reflect the way forward with Nim.

**(4)** *In the future,* if you're consuming `nim-status` as a C library (e.g. status-react on Android, iOS) and you want API compatibility with status-go, BUT you want to use the underlying Nim-only implementation, then you should compile `nim_status/c/shim.nim`  into `nim_status.a`.

Explanation: `nim_status/c/shim.nim` wraps around `nim_status/lib/*`, providing APIs that are an exact match for status-go, but status-go isn't actually involved. It's designed to be a 100% compatible drop-in replacement for status-go for users that are consuming `nim-status` as a C library.

---

What's involved in (1) - (4) above?

(1) and (2) are already done and are extremely low maintenance. Having them around makes it easier in the `nim-status` repo to e.g. run `import nim_status` and the Go shim side-by-side to test for client compatibility.

If and when we make it to v2 of nim-status, I believe that is the time to let go of the past and we can simply remove the Go shims and stuff in the Makefile related to them.

(3) will involve the bulk of our effort: interfacing with nim-waku and implementing the full functionality of a Status client. The shims should never interfere in any way with the effort to build `nim_status/lib/*.nim`.

(4) may be tough, but we shouldn't worry about it while working on (3). If we do worry about it now it will just slow us down and box us in. We can attempt to do it once (3) is far enough along. If it's not feasible then we drop (4) and come up with a plan for helping status-react to transition to the new APIs.

---

What reorganization in the code base needs to happen now to realize all of the above?

Not much.

We stop exporting the shims from the top-level `nim_status.nim`. If you want to compile or import a shim then do that directly. Top-level `nim_status.nim` should only re-export the Status client APIs developed by us in Nim. See (3) above.

We change our Makefile targets a little bit to facilitate (2) and (4). That will require only a few small easy changes in the Makefile.

We should delete `nim_status/lib/shim.nim` and `nim_status/c/lib/shim.nim` as they serve no purpose per (1) - (4) above.

We should introduce `nim_status/c/shim.nim`, and for now it should be like the `dummy.nim` in this PR.

Some of the code that's currently in `nim_status/lib.nim` is a mix of trying to be a Nim implementation and a status-go shim at the same time. We can drop, move, and/or simplify that code. The top-level `nim_status.nim` should simply re-export everything implemented in `nim_status/lib/*.nim`. Moroever, we should probably just lift `nim_status/lib/*.nim` to `nim_status/*.nim` and delete `nim_status/lib.nim`.

michaelsbradleyjr:
Closed in favor of #146, #153.

PR #148, opened 2021-01-27
experiment: creating an init() proc and openAccounts()

richard-ramos:


nim_status/lib.nim
```diff
@@ -59,3 +68,17 @@ proc identicon*(str: string): string =
...
   except:
     discard
+
+
+
+##### TODO: move to somefile
+
+type StatusObject* = object
+  accountsDB*: DbConn
+  appDB*: DbConn
```
emizzle, line 78: I feel that this name is not very descriptive of what it contains unless you think that it may contain other properties in the future. We are describing an object of database connections, so instead could we call it `DatabaseConnections` or something similar? I guess if we are planning to add other properties to the object, then maybe we should keep the name.

nim_status/lib/accounts.nim
```diff
@@ -58,51 +58,51 @@ proc test_removeDB*(accountData: string) = # TODO: remove this once proper db in
...
+    identicon* {.serializedFieldName($AccountType.Identicon), dbColumnName($AccountCol.Identicon).}: string
     keycardPairing* {.serializedFieldName($AccountType.KeycardPairing), dbColumnName($AccountCol.KeycardPairing).}: string
     keyUid* {.serializedFieldName($AccountType.KeyUid), dbColumnName($AccountCol.KeyUid).}: string
     loginTimestamp* {.serializedFieldName($AccountType.LoginTimestamp), dbColumnName($AccountCol.LoginTimestamp).}: int
 
 proc getAccounts*(db: DbConn): seq[Account] =
   var accountList: seq[Account] = @[]
-  let query = fmt"""SELECT {$AccountCol.Name}, {$AccountCol.LoginTimestamp}, {$AccountCol.PhotoPath}, {$AccountCol.KeycardPairing}, {$AccountCol.KeyUid} 
+  let query = fmt"""SELECT {$AccountCol.Name}, {$AccountCol.LoginTimestamp}, {$AccountCol.Identicon}, {$AccountCol.KeycardPairing}, {$AccountCol.KeyUid} 
   from accounts ORDER BY {$AccountCol.LoginTimestamp} DESC"""
```
emizzle, line 83:
```suggestion
  type Account* {.dbTableName("accounts").} = object
    #...

  let accounts: Accounts
  let query = fmt"""SELECT {$AccountCol.Name}, {$AccountCol.LoginTimestamp}, {$AccountCol.Identicon}, {$AccountCol.KeycardPairing}, {$AccountCol.KeyUid} 
   from {accounts.tableName} ORDER BY {$AccountCol.LoginTimestamp} DESC"""
```

emizzle (commented):
Awesome job man, looks great!

nim_status/lib.nim
richard-ramos: The name sucks but in my defense, this PR is something not mean to be merged as it is right now. It's still a WIP, but I'm pushing changes to the repo because this way it's easier to show/discuss what I'm doing

emizzle: Totally agree. No defence needed ☺️

iurimatias:
closed in favour of #158

PR #149, opened 2021-02-01
test: refactor tests so they can all be compiled and run in one pass

michaelsbradleyjr:
Make consistent use of the "test helpers" `procSuite` and `asyncTest`.

A couple of test suites are presently disabled: `accounts` and `login_and_logout`. Changes out of scope for this PR need to be introduced and the tests need to be refactored accordingly.

---

Note this PR depends on PR #146.


michaelsbradleyjr:
@richard-ramos the reason the `login_and_logout` test is disabled in this PR is explained here: https://github.com/status-im/nim-status/blob/test/test_all/test/nim/test_all.nim#L16-L20.

What do you think of moving `web3_conn` into the Status object, like you did with the database connection? I can take a crack at it, just wanted to get your thoughts first.

richard-ramos (approved):

richard-ramos:
Yeah. It makes more sense to have the web3 conn into the status object. We can do that inside the openAccounts PR once it's updated with the changes from this PR

PR #155, opened 2021-02-10
Add nim-status settings support to callPrivateRPC()

siphiuel:


michaelsbradleyjr:
Since this PR was based on the "hybrid" approach, which we're no longer pursuing for nim-status, I think we can probably just close it. Thoughts?

cc: @iurimatias

emizzle:
Closed -- hybrid approach (shim) has been abandoned.

PR #158, opened 2021-02-11
experiment: client init() proc, openAccounts(), starting waku in a separate thread

michaelsbradleyjr:
Reimplements #148 on top of the recent code reorg.


nim_status/message_thread.nim
```diff
@@ -0,0 +1,98 @@
+import
```
richard-ramos, line 1: Would message_thread.nim replace wakuV1.nim?

michaelsbradleyjr: I'm not sure. I didn't look super closely at it, just made sure all the new/changed files in #148 are in place on this PR.

michaelsbradleyjr: Ah, now I see, they're very similar but message_thread.nim has a better setup with for starting an event loop on a dedicated thread for waku. Is that correct?

PR #162, opened 2021-03-03
Increase identicon size

richard-ramos:


jrainville (approved):

michaelsbradleyjr (changes requested):
This PR branch appears to be out-of-date with respect to `master`, e.g. `identicon.nim` is in the wrong location relative to `master`.

@richard-ramos did status-go also change to generate a larger identicon?
a
If not, then these changes will break the Nim and C shim tests for identicon. That's not necessarily a problem, but we ought to then disable those tests so the CI checks pass. Also, I'm not sure of the reason but the GitHub Actions workflow didn't run for this PR.

michaelsbradleyjr:
Since nim-status is making a break from status-go, my *changes requested* comments above probably aren't relevant.

Should we keep this PR (needs rebase against `master`) or close for now and maybe do it later?

cc: @iurimatias

emizzle:
Closing due to inactivity.

Issue #167, opened 2021-03-18
list accounts


iurimatias:
proc that returns list of available accounts to login

Issue #166, opened 2021-03-18
init with config


iurimatias:
init the nim-status object given a config and setup some basic references

PR #172, opened 2021-03-19
Add StatusConfig

### LINKED ISSUE #168: initial login proc
* proc that receives account & password & decrypts/logins in that db
   * in test, simple query to certify db is now accessible

siphiuel:
closes #168  #167 #166

### DIFF
--- a/nim_status/client.nim
+++ b/nim_status/client.nim
@@ -23,7 +23,7 @@ type StatusObject* = ref object
 proc init*(config: StatusConfig): StatusObject =
   result = new StatusObject
   result.config = config
-  result.accountsDB = initializeDB($config.rootDataDir / $config.accountsDbFileName, acc_migration.newMigrationDefinition(), true) # Disabling migrations because we are reusing a status-go DB
+  result.accountsDB = initializeDB(config.rootDataDir / config.accountsDbFileName, acc_migration.newMigrationDefinition(), true) # Disabling migrations because we are reusing a status-go DB
 
 proc getAccounts*(self: StatusObject): seq[Account] =
   getAccounts(self.accountsDB)
@@ -41,9 +41,9 @@ proc getSettings*(self: StatusObject): Settings =
   getSettings(self.userDB)
 
 proc login*(self: StatusObject, keyUid: string, password: string) =
-  self.userDB = initializeDB($self.config.rootDataDir / keyUid & ".db", password, app_migration.newMigrationDefinition(), true) # Disabling migrations because we are reusing a status-go DB
+  self.userDB = initializeDB(self.config.rootDataDir / keyUid & ".db", password, app_migration.newMigrationDefinition(), true) # Disabling migrations because we are reusing a status-go DB
   echo "==============================="
-  echo "DB path: " & ($self.config.rootDataDir / keyUid & ".db")
+  echo "DB path: " & (self.config.rootDataDir / keyUid & ".db")
   echo "Password: " & password
   let result = self.userDB.value("SELECT public_key from settings") # check if decryption worked
   echo "Result: "
--- a/nim_status/client.nim
+++ b/nim_status/client.nim
@@ -2,36 +2,39 @@ import # nim libs
   os
 
 import # vendor libs
+  confutils,
+  json_serialization,
   sqlcipher
 
 import # nim-status libs
   ./accounts,
   ./chats,
+  ./config,
   ./database,
   ./migrations/sql_scripts_accounts as acc_migration,
   ./migrations/sql_scripts_app as app_migration
 
 type StatusObject* = ref object
-  rootDataDir: string
+  config*: StatusConfig
   accountsDB*: DbConn
   userDB*: DbConn
 
-proc init*(dataDir: string): StatusObject =
+proc init*(config: StatusConfig): StatusObject =
   result = new StatusObject
-  result.rootDataDir = dataDir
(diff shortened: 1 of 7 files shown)


test/client.nim
```diff
@@ -0,0 +1,65 @@
...
+
+    var account:Account = Account(
+      name: "Test",
+      loginTimestamp: 1,
+      identicon: "data:image/png;base64,something",
+      keycardPairing: "",
+      keyUid: "0x1234"
+    )
+
+    saveAccount(statusObj.accountsDB, account)
```
michaelsbradleyjr, line 25: In client we could have a:
```nim
proc saveAccount*(self: statusObj, account: Account) =
  saveAccount(self.accountsDB, account)

# usage
statusObj.saveAccount(account)
```

And similar could be done for other procs where `statusObj.accountsDB` is being passed as the first argument. It might not make sense for every such proc, but the basic idea is for the client.nim module to export an API that makes working with a client instance convenient/natural.

michaelsbradleyjr (commented):
Great start!

nim_status/client.nim
```diff
@@ -46,3 +49,8 @@ proc loadChats*(self: StatusObject): seq[Chat] =
 proc close*(self: StatusObject) =
   self.userDB.close()
   self.accountsDB.close()
+
+when isMainModule:
```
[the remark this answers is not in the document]
michaelsbradleyjr: ^ I'll take care of that, on a branch off of this PR branch.

michaelsbradleyjr:
@siphiuel, the tests should pass in GitHub Actions now. Check the WIP commit I pushed onto your PR branch, let me know if you have any questions.

siphiuel:
thanks @michaelsbradleyjr ! So the pragmas should have been executed before migrations, i get it now.

Added some more methods to StatusObject, and a line to remove the db folder once tests are executed.

test/client.nim
```diff
@@ -0,0 +1,65 @@
...
+
+    let nodeConfig = %* {"config": 1}
+    statusObj.createSettings(settingsObj, nodeConfig)
+    let dbSettings = statusObj.getSettings()
+    check:
+      dbSettings.keyUID == settingsObj.keyUID
+
+
+    statusObj.close()
+    removeDir(datadir)
```
michaelsbradleyjr, line 65: The `dataDir` is currently `test/build`, which is the same directory in which the compiled executable is located, so this line causes the executable to delete itself, which seems a bit strange, and may be problematic considering that `test/client.nim` is compiled into `test/build/test_all`, which runs more than one test that uses the `test/build` directory.

L14 could be changed to:
```nim
let dataDir = currentSourcePath.parentDir() / "build" / "data"
```

Or instead of `removeDir` you could use `removeFile` and delete the two `.db` files; see for example:
https://github.com/status-im/nim-status/blob/master/test/accounts.nim#L50

michaelsbradleyjr (changes requested):

test/client.nim
siphiuel: Thanks, fixed it. Also added an extra `createDir` call, as otherwise `confutils.load()` will complain that `Error while processing the 'rootDataDir' parameter: Directory doesn't exist`.

michaelsbradleyjr: > Also added an extra `createDir` call, as otherwise `confutils.load()` will complain that `Error while processing the 'rootDataDir' parameter: Directory doesn't exist`.

@siphiuel that's an interesting problem, hmm... it wasn't happening when `dataDir` was defined as `currentSourcePath.parentDir() / "build"`, so I'm guessing `load()` has a problem when multiple directory-levels of the path don't exist.

Take a look at waku's config: https://github.com/status-im/nim-waku/blob/master/waku/v2/node/config.nim#L81-L84.

For their `dbpath` they use `string` instead of `InputDir` or `InputFile`. So maybe our config's `rootDataDir` and `accountsDbFileName` should be `string`. That way `load()` won't try to check that the paths already exist, and we already take care of making sure the directory exists (via `createDir`) in `initializeDB` so it should be okay.

siphiuel: Right, indeed, that was caused by InputDir/InputFile doing some checking of their own. Converted to `string`.

michaelsbradleyjr (approved):

test/client.nim
```diff
@@ -0,0 +1,66 @@
...
+  chronos, confutils, json_serialization, sqlcipher, web3/conversions as web3_conversions
+
+import # nim-status libs
+  ../nim_status/[accounts, client, config, conversions, database, settings],
+  ./test_helpers
+
+procSuite "client":
+  asyncTest "client":
+
+    let dataDir = currentSourcePath.parentDir() / "build" / "data"
```
emizzle, line 14: Do we know if this will cause issues with Windows paths? @jrainville might have a suggestion. We handle similar situations in desktop like so: https://github.com/status-im/status-desktop/blob/22370e0b343da455378fd09cb322d2234241d6f8/src/app/utils/image_utils.nim#L3

emizzle (commented):
Looks great, good job! Just one comment about Windows paths in there.

test/client.nim
jrainville: How does the `/` work? Is that a standard Nim way of writing paths? I'm asking because I would have assumed you'd have to write it like this: `currentSourcePath.parentDir() & "/build/data"`

So if it's some kind of Nim function, maybe it already replaces the `/` with the right one in the OS. If so, that's pretty cool!

michaelsbradleyjr: It comes from Nim's built-in `os` module, should work fine on Windows.

PR #181, opened 2021-06-25
Fixing master

michaelsbradleyjr:
Some merges/fixes yesterday went a little off track but it wasn't apparent at the time because we were being so careful about "nuking" and checking that all tests pass.

This branch achieves the same effect as the fixes/merges that went into `master` yesterday re: remaining open PRs except for the "example client" PR, which is still in draft.

This PR exists just to check that builds and tests are working on all platforms, i.e. in GitHub Actions.


michaelsbradleyjr:
So... I just realized/remembered that GitHub Actions won't run for this PR because it has conflicts with `master`. Even changing the base to point to e.g. a branch that has all but the last commit in this branch doesn't change the situation.

Issue #183, opened 2021-06-28
Logout Feature


iurimatias:
`/logout`

logsout

michaelsbradleyjr:
Closed by #193.

PR #193, opened 2021-07-01
feat: add login and logout commands to example client

### LINKED ISSUE #182: Login Feature
- list accounts
`/list`
- press a number to select an account to select
`/login <account number>`
- type password
`<password>`

show some encrypted result

michaelsbradleyjr:
Closes #182 and #183.

### DIFF
--- a/nim_status/migrations/accounts/0001_accounts.up.sql
+++ b/nim_status/migrations/accounts/0001_accounts.up.sql
@@ -1,6 +1,7 @@
 CREATE TABLE IF NOT EXISTS accounts (
 keyUid VARCHAR PRIMARY KEY,
 name TEXT NOT NULL,
+creationTimestamp BIG INT,
 loginTimestamp BIG INT,
 photoPath TEXT,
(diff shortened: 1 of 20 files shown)


nim_status/migrations/accounts/0001_accounts.up.sql
```diff
@@ -1,6 +1,7 @@
 CREATE TABLE IF NOT EXISTS accounts (
 keyUid VARCHAR PRIMARY KEY,
 name TEXT NOT NULL,
+creationTimestamp BIG INT,
```
emizzle, line 4: Just as a general comment for this as it is something that we should revisit in the future. Instead of making schema updates in existing migrations, it may better to create a single schema sql file that we can bring up the entire DB with. Any changes to the schema can be added to the schema definition directly. Once nim-status is out in the wild, we can update existing DBs using migrations.

emizzle (approved):
Great job, just a general comment regarding DB schema and editing migration files.

PR #194, opened 2021-07-01
Import mnemonic

siphiuel:


nim_status/multiaccount.nim
```diff
@@ -49,27 +58,40 @@ proc buildAccount(privateKey: PrivateKey): Account =
 
   return acc
 
+proc deriveAddresses*(multiAcc: var MultiAccount, paths: seq[string]) =
+  for p in paths:
+    let skResult = derive(multiAcc.keySeed, KeyPath p)
+    var acc = buildAccount(cast[PrivateKey](skResult.get()))
+    acc.path = p
+    multiAcc.accounts.add(acc)
```
emizzle, line 66: What do you think about making `paths` parameter a type of `seq[KeyPath]`? 

Also, what do you think about making the `MultiAccount` param immutable (ie remove the `var`) and instead returning `seq[Account]`? Maybe it's a personal preference, but I guess I prefer not to use mutable parameters unless absolutely necessary. In that case, we are deriving accounts, not addresses, so maybe we could call it `deriveAccounts`.

In summary:
```suggestion
import
  std/sugar

proc deriveAccounts*(multiAcc: var MultiAccount, paths: seq[KeyPath]): seq[Account] =
  paths.map(path => (
      let skResult = derive(multiAcc.keySeed, KeyPath p)
      var acc = buildAccount(cast[PrivateKey](skResult.get()))
      acc.path = path
      return acc
  ))
```

nim_status/multiaccount.nim
```diff
@@ -49,27 +58,40 @@ proc buildAccount(privateKey: PrivateKey): Account =
...
-    multiAcc.mnemonic = phrase
-    multiAcc.accounts = newSeq[Account]()
-    for p in paths:
-      let skResult = derive(keySeed, KeyPath p)
-      var acc = buildAccount(cast[PrivateKey](skResult.get()))
-      acc.path = p
-      multiAcc.accounts.add(acc)
-
+    var multiAcc = importMnemonic(phrase, bip39Passphrase)
+    deriveAddresses(multiAcc,  paths)
```
emizzle, line 88: Going along with the above suggestions:
```suggestion
    var multiAcc = importMnemonic(phrase, bip39Passphrase)
    multiAcc.accounts = multiAcc.deriveAccounts(paths)
```

emizzle (commented):
Overall, awesome job, thank you so much Vitaliy! 

I hope I didn't overstep my bounds by any means with the amount of ideas/suggestions. Please take as many or as few as you'd like. 

One general comment I have is that what do you think about passing around strong types in the proc signatures, ie `KeyPath` instead of `string`?

nim_status/multiaccount.nim
```diff
@@ -49,27 +58,40 @@ proc buildAccount(privateKey: PrivateKey): Account =
...
+  if lseed < MIN_SEED_BYTES or lseed > MAX_SEED_BYTES:
+    return MultiAccount()
+
+  return MultiAccount(mnemonic: mnemonicPhrase, 
+         accounts: newSeq[Account](),
+         keyseed: seed)
+
+
 proc generateAndDeriveAddresses*(mnemonicPhraseLength: int, n: int,
   bip39Passphrase: string, paths: seq[string]): seq[MultiAccount] =
```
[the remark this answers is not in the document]
siphiuel: Yes, sure, done

nim_status/multiaccount.nim
```diff
@@ -49,27 +58,40 @@ proc buildAccount(privateKey: PrivateKey): Account =
...
   return acc
 
+proc deriveAddresses*(multiAcc: var MultiAccount, paths: seq[string]) =
+  for p in paths:
+    let skResult = derive(multiAcc.keySeed, KeyPath p)
+    var acc = buildAccount(cast[PrivateKey](skResult.get()))
+    acc.path = p
+    multiAcc.accounts.add(acc)
+
+proc importMnemonic*(mnemonicPhrase: string, bip39Passphrase: string): MultiAccount =
```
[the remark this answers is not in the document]
siphiuel: Done!

examples/chat/tui/actions.nim
```diff
@@ -128,6 +129,21 @@ proc action*(self: ChatTUI, event: ListAccountsResult) {.async, gcsafe, nimcall.
...
+# ImportMnemonicResult -----------------------------------------------------------
+
+proc action*(self: ChatTUI, event: ImportMnemonicResult) {.async, gcsafe, nimcall.} =
+  let
+    multiAcc = event.multiAcc
+    timestamp = event.timestamp
+    shouldPrint = if not self.inputReady: false else: true
+
+  # TODO need proper JSON serialization for this
+  trace "TUI importing account" #, multiAcc=(%multiAcc)
```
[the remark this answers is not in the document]
siphiuel: Fixed, thanks!

nim_status/multiaccount.nim
```diff
@@ -49,27 +58,40 @@ proc buildAccount(privateKey: PrivateKey): Account =
...
+  let seed = getSeed(Mnemonic mnemonicPhrase, bip39Passphrase)
+  # Ensure seed is within expected limits
+  let lseed = len(cast[seq[byte]](seed))
+  if lseed < MIN_SEED_BYTES or lseed > MAX_SEED_BYTES:
+    return MultiAccount()
+
+  return MultiAccount(mnemonic: mnemonicPhrase, 
+         accounts: newSeq[Account](),
+         keyseed: seed)
```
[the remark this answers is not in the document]
siphiuel: Added a separate function `importMnemonicAndDeriveAccounts()`

nim_status/multiaccount.nim
siphiuel: Done, thanks for suggestion

siphiuel: With the caveat that i kept having some weird `wrong indentation`, will look into it further.

nim_status/multiaccount.nim
siphiuel: Done

examples/chat/client.nim
```diff
@@ -54,5 +54,8 @@ proc logout*(self: ChatClient) {.async.} =
 proc generateMultiAccount*(self: ChatClient, password: string) {.async.} =
   asyncSpawn generateMultiAccount(self.taskRunner, status, password)
 
+proc importMnemonic*(self: ChatClient, mnemonic: string, passphrase: string) {.async.} =
+  asyncSpawn importMnemonic(self.taskRunner, status, mnemonic, passphrase)
```
arnetheduck, line 58: I know this is just following a pattern, but using `asyncSpawn` this way looks problematic: it starts a separate task that is disconnected from the "current" task - for example, exceptions will not be propagated and `await importMnemonic(...)` will not work as expected - instead of waiting for the import to finish, it will keep running, creating an unexpected order of operations down the line - this looks like something that needs fixing across the codebase.

examples/chat/client/tasks.nim
```diff
@@ -110,6 +110,16 @@ proc generateMultiAccount*(password: string) {.task(kind=no_rts, stoppable=false
   trace "task sent event to host", event=eventEnc, task
   asyncSpawn chanSendToHost.send(eventEnc.safe)
 
+proc importMnemonic*(mnemonic: string, passphrase: string) {.task(kind=no_rts, stoppable=false).} =
+  let 
+    multiAcc = importMnemonic(mnemonic, passphrase)
+    event = ImportMnemonicResult(multiAcc: multiAcc, timestamp: epochTime().int64)
+    eventEnc = event.encode
+    task = taskArg.taskName
```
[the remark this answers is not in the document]
siphiuel: Fixed.

examples/chat/client.nim
michaelsbradleyjr: `importMnemonic` on the RHS of `asyncSpawn` makes use of the [`{.task.}` pragma](https://github.com/status-im/nim-status/blob/feat/import-mnemonic/examples/chat/client/tasks.nim#L113). The [`task` macro](https://github.com/status-im/nim-status/blob/feat/import-mnemonic/examples/chat/task_runner/macros.nim#L13) is no work of art (it's my first "big macro", still learning), but is working as intended, which is to say it's working according to the designs/goals that @emizzle and I came up with for our [`task_runner`](https://github.com/status-im/nim-status/blob/feat/import-mnemonic/examples/chat/task_runner.nim) architecture.

In general, `{.task.}` takes the proc and rewrites it (along with some helper procs and types) so that its body runs on a [dedicated worker thread](https://github.com/status-im/nim-status/blob/feat/import-mnemonic/examples/chat/client.nim#L18) or on a thread in a pool (there are no `task_runner` pools in the example client to date, but Status Desktop does use one).

Note that `importMnemonic`, along with other tasks implemented and being implemented so far (see other open PRs in this repo), are using the `no_rts` option ("no return to sender channel/behavior"). The idea with that kind of task is that it will signal back to the host thread (generally the main thread) over an `AsyncChannel` named [`chanSendToHost`](https://github.com/status-im/nim-status/blob/feat/import-mnemonic/examples/chat/client/tasks.nim#L141). From the host thread's perspective that channel is [`chanRecvFromWorker`](https://github.com/status-im/nim-status/blob/feat/import-mnemonic/examples/chat/client/events.nim#L44). The host listens for events coming from the workers and dispatches them to [actions](https://github.com/status-im/nim-status/blob/feat/import-mnemonic/examples/chat/tui/actions.nim).

To your points about exceptions, etc.:

In the bodies of procs transformed by `{.task.}`, we can and should catch exceptions. Error-related information can then be propagated back to the main thread using `chanSendToHost`.

In some cases it will make sense to use the `rts` task option, instead of `no_rts`, which allows one to do:

```nim
let foo = await someTask(taskRunner, workerName, bar, baz)
```

A channel named `chanReturnToSender` is made available to the body of an `rts` task; when a message comes over that channel (from the worker thread), the future returned by `someTask` is automatically completed. Some thought could be given to the possibility of a convention whereby the task sends over `chanReturnToSender` a value that represents a catchable error, which then causes `someTask()` on the main thread to raise a corresponding exception.

Some general thoughts:

Our plan is to not employ `task_runner` in nim-status itself, i.e. nim-status will be single-threaded and all of its code will run in the same thread as nim-waku, making use of nim-chronos.

The example client is supposed to be a dogfooding tool for the development of nim-status, that's the main goal for it (we won't be getting too fancy with the TUI). Another goal is that by using `task_runner` to communicate from the main thread (where the TUI runs) to a worker thread running nim-status/waku, we explore and develop in an architecture that has some similarities to that of Status Desktop. As you're aware, in the latter we can't run the chronos event loop in the main thread because that's where Qt's event loop runs.

The vocabulary of the example client consists mainly of "tasks", "events", "actions", "commands". It will involve a good deal of asynchronicity and concurrency. That's a challenge, and we'll have to think carefully about how we track and maintain state, e.g. in the TUI on the main thread. But I also think it's a good thing, as it will naturally force us to think about complementary concerns re: nim-status itself.

It's early days yet for the example client; some patterns and "dos and don'ts" are still taking shape.

If you give the example client a spin, be aware that backspace/delete isn't supported yet and there are other annoyances. Those shortcomings will be remedied in the near future.

arnetheduck: task_runner or not, there are a few expectations to match when working with async code, or it becomes very difficult to reason about the code:

* _this_ `importMnemonic` advertises that it's an `async` proc that returns a `Future[void]` - when looking at such a signature, _it's expected_ that the future will complete only when the import has completed (successfully or not) - if it doesn't do that, it's breaking the expectations of the `async` framework - this can be resolved in two ways: either it should return a meaningful future, or it should not return anything and not be an `async` function at all
* `we can and should catch exceptions` - uncaught exceptions in a thread will crash the application - the way to make tasks that _don't_ have a return channel is to add `raises: [Defect]` to the generated wrapper that surrounds the transformed user code: this ensures at compile time that exceptions have been taken care of inside the task - this is likely something the `task` macro should take care of when performing the transformation in `no_rts` mode
* further, if the task does _not_ have a return channel (`no_rts`), why is a future returned, and why is `asyncSpawn` called? When is the future completed?

In general, untracked tasks like those launched by `asyncSpawn` create an unbounded queue of work - this is _sometimes_ desireable, but most of the time - it's an easy way to introduce attack surfaces, because there's no longer any backpressure on the work being done - specially since this is a model application, it's good to adhere to principles like these, so that anti-patterns don't get copy-pasted to the real thing.

nim_status/account.nim
```diff
@@ -56,16 +78,7 @@ proc child(self: ExtendedPrivKey, child: PathLevel): ExtendedPrivKeyResult =
...
-  let secretKey = hmacResult.data[0..31]
-  let chainCode = hmacResult.data[32..63]
-  let sk = SkSecretKey.fromRaw(secretKey)
-  if sk.isErr(): return err("Invalid secret key")
-
-  var extPrivK = ExtendedPrivKey(
-    secretKey: sk.get(),
-    chainCode: chainCode
-  )
+  var extPrivK = splitHMAC(cast[string](seed), masterSecret).get()
```
arnetheduck, line 81: casting to string is not safe, the in-memory representation is not the same - instead, `string.fromBytes` should be used throughout (`stew/byteutils`)

michaelsbradleyjr: Thanks, will change it!

nim_status/multiaccount.nim
```diff
@@ -49,27 +59,45 @@ proc buildAccount(privateKey: PrivateKey): Account =
...
+    let skResult = derive(multiAcc.keySeed, p)
+    var acc = buildAccount(cast[PrivateKey](skResult.get()))
+    acc.path = p
+    accounts.add(acc)
+  return accounts
+
+proc importMnemonic*(mnemonicPhrase: Mnemonic, bip39Passphrase: string): MultiAccount =
+  let seed = getSeed(Mnemonic mnemonicPhrase, bip39Passphrase)
+  # Ensure seed is within expected limits
+  let lseed = len(cast[seq[byte]](seed))
```
arnetheduck, line 74: why the cast here? it doesn't do anything - also, casting to seq should be avoided, it's generally not safe  (even if there exist a few exceptional cases)

nim_status/multiaccount.nim
```diff
@@ -49,27 +59,45 @@ proc buildAccount(privateKey: PrivateKey): Account =
 
   return acc
 
-proc generateAndDeriveAddresses*(mnemonicPhraseLength: int, n: int,
-  bip39Passphrase: string, paths: seq[string]): seq[MultiAccount] =
+proc deriveAccounts*(multiAcc: MultiAccount, paths: seq[KeyPath]): seq[Account] =
+  var accounts: seq[Account]
+  for p in paths:
+    let skResult = derive(multiAcc.keySeed, p)
+    var acc = buildAccount(cast[PrivateKey](skResult.get()))
```
arnetheduck, line 66:
```suggestion
    var acc = buildAccount(PrivateKey(skResult.get()))
```

`cast` in general is an unsafe construct that should be avoided unless you specifically know what it does and why (in which case we tend to comment why it happens to work) - when converting between distinct types, a conversion is used, like suggested

michaelsbradleyjr: All use of `cast` has been removed from modules in `nim_status`. There is still some usage in the tests, will clean that up over time.

examples/chat/client.nim
emizzle: Have added this feedback [in an issue](https://github.com/status-im/nim-status/issues/198) to be revisited and addressed.

Issue #196, opened 2021-07-05
Send Message


iurimatias:
requires #195 , to start the message should be sent to all channels

michaelsbradleyjr:
Sending a message currently works, but the topic can't be changed after starting the program, not can other topics be added removed. So all messages are sent and received re: that one topic.

The default topic is `/toy-chat/2/huilong/proto`, and it be changed by passing a cli option:
```
build/chat --waku-content-topic:"/toy-chat/2/example/proto"
```

There ended being supporting changes for this issue included in the finalized PR for the `/login` and `/logout` commands, which will be merged shortly.

Once I've added the ability to join multiple topics (#195), this issue will be unblocked and can be wrapped up quickly.

michaelsbradleyjr:
Closed by #210.

PR #210, opened 2021-07-08
feat: add join, leave, and topics commands to example client

### LINKED ISSUE #195: Join Public Chat
`/join #topic`

already works with one topic (flag at compile time), we want to make this dynamic so we can join multiple topics using nim-waku
the terminal should show messages from whatever topics we are listening to

michaelsbradleyjr:
Closes #195 and #196.

### DIFF
--- a/examples/chat.nim
+++ b/examples/chat.nim
@@ -27,3 +27,9 @@ when isMainModule:
   setControlCHook(nop)
 
   waitFor main()
+
+  # avoid exception at program exit related to a problem with a thread started
+  # by nim-eth / nim-nat-traversal sometimes not having shut down cleanly by
+  # the time chat program is ready to exit
+  # ??? does this technique work every time? are there any downsides?
+  quit(QuitSuccess)
--- a/examples/chat/client/tasks.nim
+++ b/examples/chat/client/tasks.nim
@@ -1,5 +1,5 @@
 import # std libs
-  std/[os, strutils]
+  std/[os, strutils, sets, sugar]
 
 import # nim-status libs
   ../../nim_status/[account, accounts, alias, client, database, identicon,
@@ -19,22 +19,27 @@ type
 
   StatusState* = enum loggedout, loggingin, loggedin, loggingout
 
+  WakuState* = enum stopped, starting, started, stopping
+
+const DefaultTopic = waku_chat2.DefaultTopic
+
 var
   conf {.threadvar.}: ChatConfig
   connected {.threadvar.}: bool
-  contentTopic {.threadvar.}: ContentTopic
+  contentTopics {.threadvar.}: OrderedSet[ContentTopic]
   contextArg {.threadvar.}: StatusArg
+  extIp {.threadvar.}: Option[ValidIpAddress]
+  extTcpPort {.threadvar.}: Option[Port]
+  extUdpPort {.threadvar.}: Option[Port]
+  natIsSetup {.threadvar.}: bool
   nick {.threadvar.}: string
+  nodekey {.threadvar.}: waku_chat2.crypto.PrivateKey
+  nodekeyGenerated {.threadvar.}: bool
   subscribed {.threadvar.}: bool
   status {.threadvar.}: StatusObject
   statusState {.threadvar.}: StatusState
   wakuNode {.threadvar.}: WakuNode
   wakuState {.threadvar.}: WakuState
-
-when PayloadV1:
-  var
-    symKey {.threadvar.}: SymKey
-    symKeyGenerated {.threadvar.}: bool
 
 proc resetContext() {.gcsafe, nimcall.} =
   connected = false
@@ -43,13 +48,28 @@ proc resetContext() {.gcsafe, nimcall.} =
   wakuNode = nil
   wakuState = WakuState.stopped
 
-const statusContext*: Context = proc(arg: ContextArg) {.async, gcsafe, nimcall,
+proc statusContext*(arg: ContextArg) {.async, gcsafe, nimcall,
   raises: [Defect].} =
 
-  # set threadvar values that are never reset, i.e. persist across dis/connect
+  # set threadvar values that are never reset, i.e. persist across
+  # waku chat2 dis/connect
   contextArg = cast[StatusArg](arg)
   conf = contextArg.chatConfig
-  contentTopic = conf.contentTopic
+
+  let contentTopicsStr = conf.contentTopics.strip()
+  if contentTopicsStr != "":
+    contentTopics = contentTopicsStr.split(" ").toOrderedSet()
+
+  # threadvar `natIsSetup` is a special case because the values of its
+  # counterparts `ext[Ip,TcpPort,UdpPort]` only need to be set once, i.e. they
+  # also persists across waku chat2 dis/connect; but note that
+  # `ext[Ip,TcpPort,UdpPort]` themselves are set for the first time in task
+  # `startWakuChat2` as a program startup optimization
+  natIsSetup = false
+
+  # threadvar `nodekeyGenerated` is a special case like `natIsSetup`, see
+  # previous comment
+  nodekeyGenerated = false
 
   status = StatusObject.new(conf.dataDir)
   # threadvar `statusState` is currently out of scope re: "resetting the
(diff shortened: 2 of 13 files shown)


examples/chat/client/tasks.nim
```diff
@@ -19,52 +19,65 @@ type
...
-  # set threadvar values that are never reset, i.e. persist across dis/connect
+  # set threadvar values that are never reset, i.e. persist across
+  # waku chat2 dis/connect
   contextArg = cast[StatusArg](arg)
   conf = contextArg.chatConfig
-  contentTopic = conf.contentTopic
+
+  let contentTopicsStr = conf.contentTopics.strip()
+  if contentTopicsStr != "":
+    contentTopics = contentTopicsStr.split(" ").toOrderedSet()
```
emizzle, line 61: Shouldn't we already have this available to us in `contextArg.topics`, but as an `OrderedSet[string]`? It feels like maybe the `ContextArg` type should have `OrderedSet[ContentTopic]`? But if you don't agree, then here we could simply convert from string to `ContentTopic` by way of `map`:

```nim
import sugar

contentTopics = contextArg.topics.map(t => ContentTopic t)
```

Actually, seeing that `ContentTopic` is not a distinct type of string, then likely we don't need a type conversion and can simply do:
```nim
contentTopics = contextArg.topics
```

examples/chat/client/tasks.nim
```diff
@@ -255,34 +295,36 @@ proc startWakuChat2*(username: string) {.task(kind=no_rts, stoppable=false).} =
...
+  if not nodekeyGenerated:
+    if $conf.nodekey == "":
+      nodekey = waku_chat2.crypto.PrivateKey.random(Secp256k1,
+        waku_chat2.keys.newRng()[]).tryGet()
+    else:
+      nodekey = waku_chat2.crypto.PrivateKey(scheme: Secp256k1,
+        skkey: SkPrivateKey.init(
+          waku_chat2.utils.fromHex($conf.nodekey)).tryGet())
+
+    nodekeyGenerated = true
```
emizzle, line 320: Likely out of scope for this PR, but we could send back state updates to the host thread for UI updates (ie display "Generating waku private key..." in a status bar)

examples/chat/client/tasks.nim
```diff
@@ -255,34 +295,36 @@ proc startWakuChat2*(username: string) {.task(kind=no_rts, stoppable=false).} =
...
 
   wakuNode = WakuNode.init(nodekey, ValidIpAddress.init($conf.listenAddress),
     Port(uint16(conf.tcpPort) + conf.portsShift), extIp, extTcpPort)
 
   await wakuNode.start()
 
-  wakuNode.mountRelay(conf.topics.split(" "),
-    rlnRelayEnabled = conf.rlnRelay,
+  wakuNode.mountRelay(conf.topics.split(" "), rlnRelayEnabled = conf.rlnRelay,
```
emizzle, line 327: Again, why aren't we using the topics in context?

examples/chat/client/tasks.nim
```diff
@@ -255,34 +295,36 @@ proc startWakuChat2*(username: string) {.task(kind=no_rts, stoppable=false).} =
...
-  var nodekey: waku_chat2.crypto.PrivateKey
-
-  if $conf.nodekey == "":
-    nodekey = waku_chat2.crypto.PrivateKey.random(Secp256k1,
-      waku_chat2.keys.newRng()[]).tryGet()
-  else:
-    nodekey = waku_chat2.crypto.PrivateKey(scheme: Secp256k1,
-      skkey: SkPrivateKey.init(
-        waku_chat2.utils.fromHex($conf.nodekey)).tryGet())
+  # invoke `setupNat` here instead of `statusContext` to decrease startup time
```
emizzle, line 298: The `startWakuChat2` proc is massive, and we should definitely figure out a way to split this up.

examples/chat/client/tasks.nim
```diff
@@ -346,6 +460,16 @@ proc stopWakuChat2*() {.task(kind=no_rts, stoppable=false).} =
...
+  if not wakuNode.wakuFilter.isNil():
+    let contentFilters = collect(newSeq):
+      for contentTopic in contentTopics:
+        ContentFilter(contentTopic: contentTopic)
+
+    await wakuNode.unsubscribe(FilterRequest(contentFilters: contentFilters,
+      pubSubTopic: DefaultTopic, subscribe: false))
+
+  wakuNode.unsubscribeAll(DefaultTopic)
+
```
emizzle, line 472: I'm curious why we unsubscribe when stopping waku? This is most likely due to my lack of understanding of waku.

examples/chat/client/tasks.nim
```diff
@@ -356,12 +480,28 @@ proc stopWakuChat2*() {.task(kind=no_rts, stoppable=false).} =
   trace "task sent event to host", event=eventEnc, task
   asyncSpawn chanSendToHost.send(eventEnc.safe)
 
+proc lightpushHandler(response: PushResponse) {.gcsafe.} =
+  trace "received lightpush response", response
```
emizzle, line 484: Is `lightpush` meant to facilitate notifications?

emizzle: If so, we could send an event to the host, even if we don't do anything with it right now (again likely out of scope of this PR, but just an idea).

emizzle (approved):
Really great job, well done! I have just a few questions mostly, and ideas regarding future work. All-in-all, I'd definitely say this can be merged as is and we address the ideas later on.

examples/chat.nim
```diff
@@ -27,3 +27,9 @@ when isMainModule:
   setControlCHook(nop)
 
   waitFor main()
+
+  # avoid exception at program exit related to a problem with a thread started
+  # by nim-eth / nim-nat-traversal sometimes not having shut down cleanly by
+  # the time chat program is ready to exit
+  # ??? does this technique work every time? are there any downsides?
```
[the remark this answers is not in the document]
michaelsbradleyjr: Yes, this step happens after graceful shutdown of task runner.

examples/chat/client/tasks.nim
michaelsbradleyjr: `ContentTopic` is effectively just an alias for Nim's `string`.

The threadvar here is setup as an `OrderedSet` instance for efficiency's sake.

examples/chat/client/tasks.nim
michaelsbradleyjr: Indeed.

examples/chat/client/tasks.nim
michaelsbradleyjr: waku v2 `topic`s are distinct from `contentTopic`s. See: https://rfc.vac.dev/spec/23/.

examples/chat/client/tasks.nim
michaelsbradleyjr: Yes, it should be broken up *and* this logic should really be driven by nim-status as opposed to being a sibling of nim-status logic inside `examples/chat/client/tasks.nim`.

examples/chat/client/tasks.nim
michaelsbradleyjr: That's what nim-waku's own `chat2.nim` does; it may be that it helps clean up things both locally and the remote peer, I'm not completely certain.

examples/chat/client/tasks.nim
michaelsbradleyjr: `lightpush` is as, afaik, supposed to be a mechanism for message publication that is less resource intensive than the publication via the usual "relay node" mechanism. In this case, I've simply ported nim-waku's `chat2.nim` code without deep consideration of what's involved if lightpush is active (and nim-waku's `chat2.nim` code for this case seems a bit under developed / stub'ed in, but I made a place for it to cover the bases).

examples/chat/client/tasks.nim
emizzle: Understood. More to the point of the comment was that should we consider simply using the topics already parsed in to an `OrderedSet` from the `contextArg`? Like this:
```suggestion
 contentTopics = contextArg.topics
```

examples/chat/client/tasks.nim
emizzle: Very interesting. Could we possibly use light push without running a relay node? Maybe that could be a cli option when starting up the example chat client. Something to explore another time...

Issue #199, opened 2021-07-06
Ensure generated keyUid is deterministic for new/imported accounts


emizzle:
The `keyUid` of imported accounts (via mnemonic seed) should be deterministic and match the keyUid provided by `status-go`.

michaelsbradleyjr:
Per investigation earlier today, we are generating `keyUid` deterministically, but it's not a match for `keyUid` that status-go generates for the same account.

Issue #211, opened 2021-07-12
Harden master/child keys so master cannot be leaked from child


emizzle:
There is a fair amount of "hardening" that is missing from nim-status when generating master keys. This should be added in, including unit tests.

PR #212, opened 2021-07-12
refactor(nim-status): change to a port of status-go's account generation 

emizzle:
Closes: #199

Some functions in `nim-status` were attempting to do too much in comparison to their `status-go` counterparts. The filesystem in `nim-status` varied quite a lot from the filesystem in `status-go`, which made comparing functions difficult. The end result was that certain cryptographic operations, such as generating a `keyUid` did not match between `nim-status` and `status-go`.

This PR fixes this problem by first moving some account generation files around in the filesystem to more closely mirror `status-go`. Then, the account generation functions from `status-go` were ported over to `nim-status` and the example chat client was changed to use these new functions.

Once the port from `status-go` had been completed, it could be more clearly seen that the cause of the `keyUid` mismatch was due to the use of `nim-eth`’s `PrivateKey` type over `secp256k1`’s `SkPrivateKey`. `nim-eth` does not include the `0x04` prefix that `secp256k1` uses, because `nim-eth` is more focused on ethereum implementations which do not need the prefix. Switching `PrivateKey` (in `nim-eth`) to `SkPrivateKey` (in `secp256k1`) and `PublicKey` to `SkPublicKey` allowed the `keyUid` generation to match between `nim-status` and `status-go`.

There was also some confusion between two different `Account` types in the codebase. One `Account` type was for public consumption and stored in the unencrypted accounts db. The other `Account` type contained private keys and was meant to be stored in the encrypted user db. The public `Account` type was changed to `PublicAccount` to avoid confusion.

Other notable changes include:

- chore(nim-status): Keccak 256 hash the plain text db password before keying the db.
- feat(ex-client): Move the db interactions to the nim-status client instead of doing all of it in the task.
- chore(ex-client): Remove `bip32passphrase` from `/import` arg and assume an empty string for now.

### Todo:

1. Adjust `derive` to use hardened children. This can be shown in the “derive address from imported key” test that fails when uncommented. [An issue has been created for this]([#211)


examples/chat/client/tasks.nim
```diff
@@ -89,68 +89,57 @@ proc new(T: type UserMessage, wakuMessage: WakuMessage): T =
...
-      keyUid: whisperAcct.keyUid # whisper key-uid
-    )
-
-  status.saveAccount(account)
-
-  let userDB = initializeDB(status.dataDir / account.keyUid & ".db", password)
-  userDB.close()
+    # Hardcode bip39Passphrase to empty string. Can be enabled in UI later if
+    # needed.
+    publicAccountResult = status.createAccount(12, 1, "", password,
```
michaelsbradleyjr, line 99: We need some constants here (and in other places) to help the humans.

michaelsbradleyjr (approved):
Magnificent work!!! 🔥 

This PR is massive; I've now waded through every change, but am happy to concede I might have missed something here or there.

Let's move forward by merging this and fixing consequential issues as they arise.

michaelsbradleyjr:
Because it's such a large PR, I went over the whole thing one more time.

I added a couple of small commits and intend to merge this PR once the CI lights turn green.

I didn't remove the serialization module added in `task_runner`, but it doesn't seem necessary. I disabled it and the example client still built and ran correctly. We can discuss it post-merge and remove it later (or keep it if we really need it).

PR #221, opened 2021-07-19
build: switch from MSYS2 MINGW64 to MSYS2 UCRT64

michaelsbradleyjr:
Gives better compatibility at compile and runtime, and notably better performance re: e.g. disk IO.


emizzle (approved):
Looks good to me (which isn't saying much 😆)

Issue #214, opened 2021-07-12
List Wallet Accounts


iurimatias:
`/wallet-list`
> list wallet accounts

Issue #197, opened 2021-07-05
Generate new wallet account given private key


iurimatias:
`/addaccount <privatekey>`

should add account to the logged in user, in a row in the database

emizzle:
Review need for weakened db encryption. There may be no longer be a need, unless if nim-status is used in mobile in the future, at which point we might be able to add in with a compilation flag or similar.
https://github.com/status-im/nim-status/blob/master/nim_status/database.nim#L28-L31
```
result.exec("PRAGMA cipher_page_size = 1024")
  result.exec("PRAGMA kdf_iter = 3200")
  result.exec("PRAGMA cipher_hmac_algorithm = HMAC_SHA1")
  result.exec("PRAGMA cipher_kdf_algorithm = PBKDF2_HMAC_SHA1")
  result.exec("PRAGMA foreign_keys = ON")
  result.exec("PRAGMA journal_mode = WAL")
```

PR #223, opened 2021-07-19
feat: generate wallet account

emizzle:
Closes: #214, #215.

Almost all the work for #197 is included in this PR.

Generate a new wallet account using `/add`, `/addaccount`

feat: create settings when creating/importing a login account
When creating or importing a new account (a login account, not a wallet account), settings are now created.

feat: list wallet accounts
Wallet accounts can be listed using `/wallets`, `/listwalletaccounts`, and `/listwallets`. This gets all accounts in the user db that are not chat accounts. It would be much better to store all wallet accounts in the db using the `wallet` flag, however status-go did not do this, reserving that flag for the root wallet account and creating a unique db constraint on that column.


michaelsbradleyjr (changes requested):
Another massive effort, very well done!

I've marked it with *Request changes* only because I want to understand better whey the auto-connect after login (in example client) is disabled, and determine if the empty topic indicator in the TUI is a result of some change in this PR, or is owing to something outside this PR's scope, e.g. some change in the waku network.

Finally, I note that it seems like `nim_status/client.nim` is getting pretty big, maybe there's a way to break it down a bit. Or maybe not, and I don't think it's a blocker for this PR.

examples/chat/tui/actions.nim
```diff
@@ -202,8 +261,8 @@ proc action*(self: ChatTUI, event: LoginResult) {.async, gcsafe, nimcall.} =
   else:
     self.client.account = event.account
     self.printResult("Login successful.", getTime().toUnix())
-    if not self.client.online:
-      asyncSpawn self.client.connect(self.client.account.name)
+    # if not self.client.online:
```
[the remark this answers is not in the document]
emizzle: Re-enabled. As we discussed, the missing content topic may be caused by an issue with the waku network.

examples/chat/tui/actions.nim
```diff
@@ -93,6 +96,25 @@ proc action*(self: ChatTUI, event: InputString) {.async, gcsafe, nimcall.} =
...
+      self.wprintFormatError(event.timestamp, event.error)
+      return
+
+    let
+      timestamp = event.timestamp
+      name = event.name
+      address = $event.address
+
+    self.printResult("Added wallet account:", timestamp)
+    self.printResult(fmt"{2.indent()}{name} ({address})", timestamp)
```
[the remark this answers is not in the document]
emizzle: I kind of feel like there should be a way to see the entire address so that a user can double-check the address before executing any transactions. What I have done is show the abbreviated version upon wallet creation, but kept the long form for listing.

examples/chat/client/tasks.nim
```diff
@@ -98,20 +101,75 @@ proc new(T: type UserMessage, wakuMessage: WakuMessage): T =
...
+  trace "task sent event to host", event=eventEnc, task
+  asyncSpawn chanSendToHost.send(eventEnc.safe)
 
 proc createAccount*(password: string) {.task(kind=no_rts, stoppable=false).} =
+  let timestamp = getTime().toUnix
+
+  if statusState != StatusState.loggedout:
+    let
+      eventNotLoggedOut = AddWalletAccountResult(error: "You must be logged " &
+        "out to create a new account. Please logout with /logout first.",
```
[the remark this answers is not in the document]
emizzle: Fixed.

examples/chat/client/tasks.nim
```diff
@@ -98,20 +101,75 @@ proc new(T: type UserMessage, wakuMessage: WakuMessage): T =
...
+
+proc addWalletAccount*(name: string,
+  password: string) {.task(kind=no_rts, stoppable=false).} =
+
+  let timestamp = getTime().toUnix
+
+  if statusState != StatusState.loggedin:
+    let
+      eventNotLoggedIn = AddWalletAccountResult(error: "Not logged in, " &
+        "cannot create a new wallet account. Please login with /login first.",
```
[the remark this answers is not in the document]
emizzle: Fixed.

examples/chat/client/tasks.nim
```diff
@@ -210,35 +290,44 @@ proc login*(account: int, password: string) {.task(kind=no_rts, stoppable=false)
   let allAccounts = status.getPublicAccounts()
 
   var
-    event: LoginResult
+    event: events.LoginResult
```
siphiuel, line 293: Noticed the `events` prefix - does it make it sense to enforce name qualification via `from ./events import nil` instead of `import ./events`?

emizzle: What would be the impact on the end result?

siphiuel: Possibly preventing name collisions in the future.

emizzle: Yeah, i agree that using the suffix `Result` for events is a bit misleading/confusing since we are now heavily using the `nim-result` lib which uses types that end in `Result`.

nim_status/accounts/generator/generator.nim
```diff
@@ -64,9 +62,9 @@ proc deriveChildAccount(self: Generator, a: Account,
 proc deriveChildAccounts(self: Generator, a: Account,
   paths: seq[KeyPath]): DeriveChildAccountsResult =
 
-  var derived: seq[Account] = @[]
+  var derived = initTable[KeyPath, Account]()
   for path in paths:
-    derived.add ?self.deriveChildAccount(a, path)
+    derived[path] = ?self.deriveChildAccount(a, path)
 
   ok(derived)
```
siphiuel, line 69: Super-minor comment: not related to this PR, but sometimes we use unqualified `Result.ok`, and sometimes qualified.

nim_status/client.nim
```diff
@@ -90,58 +153,228 @@ proc storeDerivedAccountsInDb(self: StatusObject, id: UUID, keyUid: string,
...
 
-proc createAccount*(self: StatusObject,
-  mnemonicPhraseLength, n: int, bip39Passphrase, password: string,
-  paths: seq[KeyPath], dir: string): PublicAccountResult =
+    return PublicAccountResult.ok(pubAccount)
+  except Exception as e:
+    return PublicAccountResult.err e.msg
+
+proc addWalletAccount*(self: StatusObject, name, password,
+  #[privateKey: Option[string],]# dir: string): WalletAccountResult =
```
siphiuel, line 182: Quite ignorant about it - does this mean we might possibly support specifying private keys for wallet account creation?

nim_status/accounts/generator/generator.nim
emizzle: Very relevant. This was a leftover from previous code. I will update it.

nim_status/client.nim
emizzle: Yes, it is something I am working on now. I have decided against including the private key as a parameter and will create a new proc for it.

nim_status/client.nim
```diff
@@ -90,58 +153,228 @@ proc storeDerivedAccountsInDb(self: StatusObject, id: UUID, keyUid: string,
...
+  let
+    pubAccount = ?self.storeDerivedAccounts(account.id, account.keyUid, paths,
+      password, dir)
+    uuidResult = uuidGenerate()
+
+  if uuidResult.isErr:
+    return PublicAccountResult.err "Error generating uuid: " & $uuidResult.error
+  
+  let 
+    settings = Settings(
```
siphiuel, line 232: Possible refactor - move settings creation to a separate proc?

nim_status/client.nim
```diff
@@ -90,58 +153,228 @@ proc storeDerivedAccountsInDb(self: StatusObject, id: UUID, keyUid: string,
...
+    return GetSettingsResult.err e.msg
+
+proc importMnemonic*(self: StatusObject, mnemonic: Mnemonic,
+  bip39Passphrase, password: string, dir: string): PublicAccountResult =
+
+  try:
+    let
+      imported = ?self.accountsGenerator.importMnemonic(mnemonic,
+        bip39Passphrase)
+      paths = @[PATH_WALLET_ROOT, PATH_EIP_1581, PATH_WHISPER,
```
siphiuel, line 340: These paths are also declared in `createAccount`, maybe makes sense to extract it into a `const` or something.

siphiuel (approved):
A formidable PR! 👍  Left a couple of minor comments.

michaelsbradleyjr (approved):

michaelsbradleyjr:
I added one commit, restoring `topic` to the constructor call for `UserMessage`.

nim_status/accounts/generator/generator.nim
emizzle: Fixed.

examples/chat/client/tasks.nim
emizzle: All `Result` suffixes for event types have been replaced with `Event`.

nim_status/client.nim
emizzle: Hmm all the paths are already constants, but i do see how these four paths are used together in two spots. I feel like pulling more out in to constants isn't really going to benefit us too much here. I'd prefer to leave it as is.

emizzle:
All comments have been addressed, this should be ready to go.

PR #229, opened 2021-07-21
feat: support importing wallet accout via private key

emizzle:
Closes: #197 

Add funcs to support importing a wallet account via private key.

Support password validation before importing the keystore file. Prevent duplicate keystore files from being imported.

Add keystore file unit tests for storing and loading of key files

Add import wallet pk command to example chat client

Refactor web3_conversions imports to insteaad import `nim_status/conversions`.

### DIFF
--- a/nim_status/accounts/generator/generator.nim
+++ b/nim_status/accounts/generator/generator.nim
@@ -1,18 +1,21 @@
 import # std libs
-  std/[json, os, strformat, strutils, tables]
+  std/[json, os, strformat, strutils, sugar, tables, times]
 
 import # vendor libs
   eth/keys, eth/keyfile/[keyfile, uuid], json_serialization, secp256k1,
-  stew/results
+  stew/results, web3/ethtypes
 
 import # nim-status libs
-  ./account, ../../extkeys/[hdkey, mnemonic, types], ../public_accounts, ./utils
+  ./account, ../../conversions, ../../extkeys/[hdkey, mnemonic, types],
+  ../public_accounts, ./utils
 
 export utils
 
 type
   Generator* = ref object
-    accounts*: TableRef[string, Account]
+    accounts*: TableRef[string, Account] # TODO: this should be made private,
+      # and we should adjust tests accordingly. The tests should reflect client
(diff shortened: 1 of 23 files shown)


michaelsbradleyjr (approved):

nim_status/accounts/generator/generator.nim
```diff
@@ -175,37 +203,57 @@ proc reset(self: Generator) =
...
+    version, cryptkind, kdfkind, workfactorFinal)
+
+  if keyFileJsonResult.isErr:
+    return StoreKeyFileResult.err "Error creating key file: " &
+      $keyFileJsonResult.error
+
+  let
+    keyFileJson = $keyFileJsonResult.get
+    now {.used.} = now().format("yyyy-MM-dd'T'HH-mm-ss'.'fffffffff'Z'")
+    keyFilePath = dir / fmt"UTC--{now}--{($address).strip0xPrefix}"
```
michaelsbradleyjr, line 233: There's one thing I noticed when trying out the branch (after my earlier PR review/approval):

The format has UTC but `now()` is giving us the local time. For example, a keystore that was created by the example client a few minutes ago has the filename `UTC--2021-07-21T12-24-44.586918000Z--...`, where `12-24-44...` is the local time for me.

We should instead shift the datetime to actually be UTC. I'll figure out how to do it and push a commit.

michaelsbradleyjr (changes requested):

michaelsbradleyjr (approved):
I pushed a commit changing `now()` to `now().utc` in the context of generating keystore files.

There are some other places where we use `now()`. For example: https://github.com/status-im/nim-status/blob/feat/import-wallet-pk/nim_status/accounts/accounts.nim#L48-L51.

I opened the database with *DB Browser for SQLite* and did see that those timestamps are made w.r.t. the local timezone. Maybe we should also change those to be UTC?

nim_status/accounts/generator/generator.nim
michaelsbradleyjr: See: https://github.com/status-im/nim-status/pull/229#pullrequestreview-712031157.

PR #233, opened 2021-07-22
feat: import wallet account via mnemonic

emizzle:
Closes: #224.

Allows a user to import a wallet account via 12-word mnemonic seed.

The account is first imported in the account generator. Then a wallet account is derived at the root wallet path (m/44'/60’/0’/0/0) from the master key of the mnemonic. This account is then stored in the Accounts table of the user’s db.

### DIFF
--- a/examples/chat/client/tasks.nim
+++ b/examples/chat/client/tasks.nim
@@ -138,7 +138,7 @@ proc addWalletAccount*(name: string,
 
   let
     walletAccount = walletAccountResult.get
-    walletName = if walletAccount.name.isNone: "" else: walletAccount.name.get
+    walletName = walletAccount.name.get("")
     event = AddWalletAccountEvent(name: walletName,
       address: walletAccount.address, timestamp: timestamp)
     eventEnc = event.encode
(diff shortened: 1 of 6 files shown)


examples/chat/client/tasks.nim
```diff
@@ -190,6 +190,49 @@ proc addWalletPrivateKey*(name: string, privateKey: string, password: string)
   trace "task sent event to host", event=eventEnc, task
   asyncSpawn chanSendToHost.send(eventEnc.safe)
 
+proc addWalletSeed*(name: string, mnemonic: string, password: string,
```
siphiuel, line 193: Styling: consider removing superfluous string type specifiers

michaelsbradleyjr: Like this?
```nim
proc addWalletSeed*(name, mnemonic, password, bip39Passphrase: string)
```

michaelsbradleyjr (approved):

examples/chat/client/tasks.nim
```diff
@@ -190,6 +190,49 @@ proc addWalletPrivateKey*(name: string, privateKey: string, password: string)
...
+        "error: " & walletAccountResult.error, timestamp: timestamp)
+      eventEnc = event.encode
+      task = taskArg.taskName
+    trace "task sent event with error to host", event=eventEnc, task
+    asyncSpawn chanSendToHost.send(eventEnc.safe)
+    return
+
+  let
+    walletAccount = walletAccountResult.get
+    walletName = if walletAccount.name.isNone: "" else: walletAccount.name.get
```
siphiuel, line 227: There is an overloaded `Option.get` with a default parameter: https://nim-lang.org/docs/options.html#get,Option[T],T

siphiuel (approved):
Nice!

examples/chat/client/tasks.nim
siphiuel: Yes!

michaelsbradleyjr: Cool, I just wanted to make sure I understood you correctly. I often forget to take advantage of that shortcut, thanks for the tip!

emizzle: After trying to make this change, I remember now why I didn't do this perviously for tasks. When combining same-typed parameters in the signature *of a task proc*, the compiler barfs with this error:
```
/Users/emizzle/repos/status-im/nim-status/examples/chat/client/tasks.nim(193, 37) Error: initialization not allowed here
```
I really don't know why, but I'm assuming it has something to do with the `task` macro.

examples/chat/client/tasks.nim
emizzle: This is awesome, great look!

PR #234, opened 2021-07-22
feat: Add watch-only wallet account

emizzle:
Closes: #226

Add command (`/addwalletwatch`, `/addwatch`) that allows adding a watch-only account to the example client.

### DIFF
--- a/examples/chat/client/tasks.nim
+++ b/examples/chat/client/tasks.nim
@@ -233,6 +233,59 @@ proc addWalletSeed*(name: string, mnemonic: string, password: string,
   trace "task sent event to host", event=eventEnc, task
   asyncSpawn chanSendToHost.send(eventEnc.safe)
 
+proc addWalletWatchOnly*(address: string,
+  name: string) {.task(kind=no_rts, stoppable=false).} =
+
+  let timestamp = getTime().toUnix
+
+  if statusState != StatusState.loggedin:
+    let
+      eventNotLoggedIn = AddWalletAccountEvent(error: "Not logged in, " &
+        "cannot add a new wallet account.",
+        timestamp: timestamp)
+      eventNotLoggedInEnc = eventNotLoggedIn.encode
+      task = taskArg.taskName
+
+    trace "task sent event to host", event=eventNotLoggedInEnc, task
+    asyncSpawn chanSendToHost.send(eventNotLoggedInEnc.safe)
+    return
+
+  var addressParsed: Address
+  try:
+    addressParsed = address.parseAddress
+  except Exception as e:
+    let
+      event = AddWalletAccountEvent(error: "Error adding watch-only wallet " &
+        "account: " & e.msg, timestamp: timestamp)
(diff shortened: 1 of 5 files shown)


examples/chat/client/tasks.nim
```diff
@@ -233,6 +233,58 @@ proc addWalletSeed*(name: string, mnemonic: string, password: string,
...
+  try:
+    addressParsed = address.parseAddress
+  except Exception as e:
+    let
+      event = AddWalletAccountEvent(error: "Error adding watch-only wallet " &
+        "account: " & e.msg, timestamp: timestamp)
+      eventEnc = event.encode
+      task = taskArg.taskName
+    trace "task sent event with error to host", event=eventEnc, task
+    asyncSpawn chanSendToHost.send(eventEnc.safe)
```
michaelsbradleyjr, line 263: Should there be an early `return` here?

nim_status/client.nim
```diff
@@ -274,6 +274,16 @@ proc addWalletSeed*(self: StatusObject, mnemonic: Mnemonic, name, password,
   except Exception as e:
     return WalletAccountResult.err e.msg
 
+proc addWalletWatchOnly*(self: StatusObject, address: Address,
+  name: string): WalletAccountResult =
+
+  try:
+    return self.storeWalletAccount(name, address, SkPublicKey.none,
```
michaelsbradleyjr, line 281: Is `return` necessary here and on L286? It seems like the expressions being of type `WalletAccountResult` should suffice for them to be returned automatically, but maybe that doesn't work in this kind of setup, not sure.

michaelsbradleyjr (changes requested):

examples/chat/client/tasks.nim
```diff
@@ -233,6 +233,58 @@ proc addWalletSeed*(name: string, mnemonic: string, password: string,
...
+        "account: " & walletAccountResult.error, timestamp: timestamp)
+      eventEnc = event.encode
+      task = taskArg.taskName
+    trace "task sent event with error to host", event=eventEnc, task
+    asyncSpawn chanSendToHost.send(eventEnc.safe)
+    return
+
+  let
+    walletAccount = walletAccountResult.get
+    walletName = if walletAccount.name.isNone: "" else: walletAccount.name.get
```
michaelsbradleyjr, line 279: See @siphiuel's tip on your other open PR: https://github.com/status-im/nim-status/pull/233#discussion_r674870092.

emizzle: Updated.

examples/chat/client/tasks.nim
emizzle: Great catch, updated.

nim_status/client.nim
emizzle: It is not, was simply to be explicit for readability's sake.

michaelsbradleyjr (approved):

PR #237, opened 2021-07-22
Custom tokens support in example chat app

### LINKED ISSUE #227: List Known Tokens

### LINKED ISSUE #228: Add/Remove Custom Token

siphiuel:
Fixes #227
Fixes #228

### DIFF
--- a/examples/chat/client/tasks.nim
+++ b/examples/chat/client/tasks.nim
@@ -782,8 +782,11 @@ proc addCustomToken*(address: string, name: string, symbol: string, color: strin
     return
 
   try:
-    let
-      addResult = status.addCustomToken(address.parseAddress, name, symbol, color, decimals.parseUInt)
+    var uintDecimals: uint
+    if decimals != "":
+      uintDecimals = decimals.parseUInt
+
+    let addResult = status.addCustomToken(address.parseAddress, name, symbol, color, uintDecimals)
 
     if addResult.isErr:
       raise newException(Exception, addResult.error)
--- a/examples/chat/client/tasks.nim
+++ b/examples/chat/client/tasks.nim
@@ -811,7 +811,7 @@ proc addCustomToken*(address: string, name: string, symbol: string, color: strin
     asyncSpawn chanSendToHost.send(eventEnc.safe)
 
 
-proc deleteCustomToken*(address: string) {.task(kind=no_rts, stoppable=false).} =
+proc deleteCustomToken*(indexStr: string) {.task(kind=no_rts, stoppable=false).} =
   let timestamp = getTime().toUnix
 
   if statusState != StatusState.loggedin:
@@ -827,14 +827,27 @@ proc deleteCustomToken*(address: string) {.task(kind=no_rts, stoppable=false).} 
     return
 
   try:
+    let 
+      index = indexStr.parseInt
+      allTokens = status.getCustomTokens().get
+    if index < 1 or index > allTokens.len:
+      let
+        event = DeleteCustomTokenEvent(error: "bad token number")
+        eventEnc = event.encode
+        task = taskArg.taskName
+      trace "task sent event with error to host", event=eventEnc, task
+      asyncSpawn chanSendToHost.send(eventEnc.safe)
+      return
+
     let
-      deleteResult = status.deleteCustomToken(address.parseAddress) 
+      address = allTokens[index - 1].address
+      deleteResult = status.deleteCustomToken(address) 
 
     if deleteResult.isErr:
       raise newException(Exception, deleteResult.error)
 
     let
-      event = DeleteCustomTokenEvent(address: address, timestamp: timestamp)
+      event = DeleteCustomTokenEvent(address: $address, timestamp: timestamp)
       eventEnc = event.encode
       task = taskArg.taskName
 
@@ -842,7 +855,7 @@ proc deleteCustomToken*(address: string) {.task(kind=no_rts, stoppable=false).} 
     asyncSpawn chanSendToHost.send(eventEnc.safe)
   except Exception as e:
     let
-      event = DeleteCustomTokenEvent(error: "Error deleting a custom token, " &
+      event = DeleteCustomTokenEvent(error: "Error deleting custom token, " &
         "error: " & e.msg, timestamp: timestamp)
       eventEnc = event.encode
       task = taskArg.taskName
--- a/examples/chat/client/tasks.nim
+++ b/examples/chat/client/tasks.nim
@@ -731,6 +731,21 @@ proc publishWakuChat2*(message: string) {.task(kind=no_rts, stoppable=false).} =
         asyncSpawn wakuNode.publish(DefaultTopic, wakuMessage, conf.rlnRelay)
 
 proc getCustomTokens*() {.task(kind=no_rts, stoppable=false).} =
+  let timestamp = getTime().toUnix
+
+  if statusState != StatusState.loggedin:
+    let
+      eventNotLoggedIn = GetCustomTokensEvent(error: "Not logged in, " &
+        "cannot get custom tokens.",
+        timestamp: timestamp)
+      eventNotLoggedInEnc = eventNotLoggedIn.encode
+      task = taskArg.taskName
+
+    trace "task sent event to host", event=eventNotLoggedInEnc, task
+    asyncSpawn chanSendToHost.send(eventNotLoggedInEnc.safe)
+    return
+
+
   let tokens = status.getCustomTokens()
   if tokens.isErr:
     let
@@ -766,29 +781,32 @@ proc addCustomToken*(address: string, name: string, symbol: string, color: strin
     asyncSpawn chanSendToHost.send(eventNotLoggedInEnc.safe)
     return
 
-  let
-    addResult = status.addCustomToken(address, name, symbol, color, decimals) 
+  try:
+    let
+      addResult = status.addCustomToken(address.parseAddress, name, symbol, color, decimals.parseUInt)
 
-  if addResult.isErr:
+    if addResult.isErr:
+      raise newException(Exception, addResult.error)
+
+    let
+      token = addResult.get
+      event = AddCustomTokenEvent(address: $token.address,
+        name: token.name, symbol: token.symbol, color: token.color, 
+        decimals: token.decimals, timestamp: timestamp)
+      eventEnc = event.encode
+      task = taskArg.taskName
+
+    trace "task sent event to host", event=eventEnc, task
+    asyncSpawn chanSendToHost.send(eventEnc.safe)
+  except Exception as e:
     let
       event = AddCustomTokenEvent(error: "Error adding a custom token, " &
-        "error: " & addResult.error, timestamp: timestamp)
+        "error: " & e.msg, timestamp: timestamp)
       eventEnc = event.encode
       task = taskArg.taskName
     trace "task sent event with error to host", event=eventEnc, task
     asyncSpawn chanSendToHost.send(eventEnc.safe)
-    return
 
-  let
-    token = addResult.get
-    event = AddCustomTokenEvent(address: $token.address,
-      name: token.name, symbol: token.symbol, color: token.color, 
-      decimals: token.decimals, timestamp: timestamp)
-    eventEnc = event.encode
-    task = taskArg.taskName
-
-  trace "task sent event to host", event=eventEnc, task
-  asyncSpawn chanSendToHost.send(eventEnc.safe)
 
 proc deleteCustomToken*(address: string) {.task(kind=no_rts, stoppable=false).} =
   let timestamp = getTime().toUnix
@@ -805,23 +823,26 @@ proc deleteCustomToken*(address: string) {.task(kind=no_rts, stoppable=false).} 
     asyncSpawn chanSendToHost.send(eventNotLoggedInEnc.safe)
     return
 
-  let
-    deleteResult = status.deleteCustomToken(address) 
+  try:
+    let
+      deleteResult = status.deleteCustomToken(address.parseAddress) 
 
-  if deleteResult.isErr:
+    if deleteResult.isErr:
+      raise newException(Exception, deleteResult.error)
+
+    let
+      event = DeleteCustomTokenEvent(address: address, timestamp: timestamp)
+      eventEnc = event.encode
+      task = taskArg.taskName
+
+    trace "task sent event to host", event=eventEnc, task
+    asyncSpawn chanSendToHost.send(eventEnc.safe)
+  except Exception as e:
     let
       event = DeleteCustomTokenEvent(error: "Error deleting a custom token, " &
-        "error: " & deleteResult.error, timestamp: timestamp)
+        "error: " & e.msg, timestamp: timestamp)
       eventEnc = event.encode
       task = taskArg.taskName
     trace "task sent event with error to host", event=eventEnc, task
     asyncSpawn chanSendToHost.send(eventEnc.safe)
-    return
 
-  let
-    event = DeleteCustomTokenEvent(address: address, timestamp: timestamp)
-    eventEnc = event.encode
-    task = taskArg.taskName
-
-  trace "task sent event to host", event=eventEnc, task
-  asyncSpawn chanSendToHost.send(eventEnc.safe)
--- a/examples/chat/client/tasks.nim
+++ b/examples/chat/client/tasks.nim
@@ -865,7 +865,7 @@ proc deleteCustomToken*(index: int) {.task(kind=no_rts, stoppable=false).} =
 
       trace "task sent event to host", event=eventEnc, task
       asyncSpawn chanSendToHost.send(eventEnc.safe)
-  except Exception as e:
+  except CatchableError as e:
     let
       event = DeleteCustomTokenEvent(error: "Error deleting custom token, " &
         "error: " & e.msg, timestamp: timestamp)
--- a/examples/chat/client/tasks.nim
+++ b/examples/chat/client/tasks.nim
@@ -1,5 +1,7 @@
 import # std libs
-  std/[os, strutils, sets, sugar, times]
+  std/[os, strutils, sets, times]
+
+from std/sugar import `=>`, collect
 
 import # nim-status libs
   ../../nim_status/[conversions, client, database],
@@ -766,7 +768,7 @@ proc getCustomTokens*() {.task(kind=no_rts, stoppable=false).} =
   trace "task sent event to host", event=eventEnc, task
   asyncSpawn chanSendToHost.send(eventEnc.safe)
 
-proc addCustomToken*(address: string, name: string, symbol: string, color: string, decimals: string) {.task(kind=no_rts, stoppable=false).} =
+proc addCustomToken*(address: Address, name: string, symbol: string, color: string, decimals: uint) {.task(kind=no_rts, stoppable=false).} =
   let timestamp = getTime().toUnix
 
   if statusState != StatusState.loggedin:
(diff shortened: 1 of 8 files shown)


examples/chat/tui/commands.nim
```diff
@@ -27,6 +27,62 @@ const hashCharSet = {'#'}
...
+    if command.address == "":
+      self.wprintFormatError(epochTime().int64,
+        "address cannot be empty, please provide an address.")
+    elif command.name == "":
+      self.wprintFormatError(epochTime().int64,
+        "name cannot be empty, please provide a name.")
+    elif command.symbol == "":
+      self.wprintFormatError(epochTime().int64,
+        "symbol cannot be empty, please provide a symbol.")
+    else:
```
emizzle, line 80: Because there are two optional parameters, we will never be able to know if the user was specifying `color` and not `decimals` or `decimals` and not `color`. So maybe we should provide some validation here, along the lines of "if color specified, check for valid hex string. if decimals specified, check for valid int". What do you think?

examples/chat/tui/commands.nim
```diff
@@ -27,6 +27,62 @@ const hashCharSet = {'#'}
...
+      asyncSpawn self.client.addCustomToken(command.address, command.name, command.symbol, command.color, command.decimals)
+  except:
+    self.wprintFormatError(epochTime().int64, "invalid arguments.")
```
emizzle, line 83: Will the source of error always be invalid arguments? Maybe we could catch the exception and display the error message, like so:
```suggestion
  except Exception as e:
    self.wprintFormatError(epochTime().int64, "Error adding custom token: " & e.msg)
```

examples/chat/tui/commands.nim
```diff
@@ -157,6 +213,38 @@ proc command*(self: ChatTUI, command: CreateAccount) {.async, gcsafe,
...
+  nimcall.} =
+
+  try:
+    if command.address == "":
+      self.wprintFormatError(epochTime().int64,
+        "address cannot be empty, please provide an address.")
+    else:
+      asyncSpawn self.client.deleteCustomToken(command.address)
+  except:
+    self.wprintFormatError(epochTime().int64, "invalid arguments.")
```
emizzle, line 246: Same suggestion as above (invalid arguments).

examples/chat/tui/commands.nim
```diff
@@ -175,6 +263,23 @@ proc command*(self: ChatTUI, command: Disconnect) {.async, gcsafe, nimcall.} =
...
+    "all existing custom tokens.")
+
+proc new*(T: type GetCustomTokens, args: varargs[string]): T =
+  T()
+
+proc split*(T: type GetCustomTokens, argsRaw: string): seq[string] =
+  @[]
+
+proc command*(self: ChatTUI, command: GetCustomTokens) {.async, gcsafe, nimcall.} =
+  asyncSpawn self.client.getCustomTokens()
```
emizzle, line 280: Do we need a try/except surrounding the `getCustomTokens` call?

nim_status/client.nim
```diff
@@ -224,6 +230,36 @@ proc storeImportedWalletAccount(self: StatusObject, privateKey: SkSecretKey,
...
+proc addCustomToken*(self: StatusObject, address, name, symbol, color, decimals: string): AddCustomTokenResult =
+  if not self.isLoggedIn:
+    return AddCustomTokenResult.err "Not logged in. You must be logged in to " &
+      "add a new custom token."
+
+  var uintDecimals: uint
+  if decimals != "":
+    uintDecimals = parseUInt(decimals)
+  # TODO
+  let token = Token(address: cast[Address](address), name: name, symbol: symbol, color: color, decimals: uintDecimals)
```
emizzle, line 242: To me, it makes more sense to accept typed arguments (`address: Address` and `decimals: int`) in the `addCustomToken` proc signature as we are operating in the nim-status library which requires these types to function correctly. It should be up to the consuming library to convert to the correct type before calling functions in `nim-status/client`.
```suggestion
  let token = Token(address: address, name: name, symbol: symbol, color: color, decimals: uintDecimals)
```
Note also, that types should be *type converted* instead of *casted* where possible, ie `Address(address)` instead of `cast[Address](address)` as a cast is not needed in the majority of cases and could cause unwanted side effects if not used correctly.

nim_status/client.nim
```diff
@@ -224,6 +230,36 @@ proc storeImportedWalletAccount(self: StatusObject, privateKey: SkSecretKey,
...
+  try:
+    self.userDb.addCustomToken(token)
+  except Exception as e:
+    return AddCustomTokenResult.err "Error adding a custom token: " & e.msg
+
+  AddCustomTokenResult.ok(token)
+
+proc deleteCustomToken*(self: StatusObject, address: string): DeleteCustomTokenResult =
```
emizzle, line 250: Here too, I think it would be better to accept a type of `Address`:
```suggestion
proc deleteCustomToken*(self: StatusObject, address: Address): DeleteCustomTokenResult =
```

examples/chat/client/tasks.nim
```diff
@@ -633,3 +633,99 @@ proc publishWakuChat2*(message: string) {.task(kind=no_rts, stoppable=false).} =
...
+  let
+    token = addResult.get
+    event = AddCustomTokenEvent(address: $token.address,
+      name: token.name, symbol: token.symbol, color: token.color, 
+      decimals: token.decimals, timestamp: timestamp)
+    eventEnc = event.encode
+    task = taskArg.taskName
+
+  trace "task sent event to host", event=eventEnc, task
+  asyncSpawn chanSendToHost.send(eventEnc.safe)
```
emizzle, line 695: See comments below, but I think we should be converting types here as satisfying the type needs of the nim-status lib:
```suggestion
  import ../../nim-status/conversions
  
  try: # needed for parseAddress and parseInt, and can also be used to catch any errors with addCustomToken
    let addResult = status.addCustomToken(address.parseAddress, name, symbol, color, decimals.parseInt)
    
    # do the rest...
    
    except Exception as e:
      let
        eventError = AddCustomTokenEvent(error: e.msg, # prob don't wanna add more text, else when the error bubbles, the text will compound
          timestamp: timestamp)
        eventErrorEnc = eventNotLoggedIn.encode
        task = taskArg.taskName

      trace "task sent error to host", event=eventErrorEnc, task
      asyncSpawn chanSendToHost.send(eventErrorEnc.safe)
      return
  
```

examples/chat/client/tasks.nim
```diff
@@ -633,3 +633,99 @@ proc publishWakuChat2*(message: string) {.task(kind=no_rts, stoppable=false).} =
           lightpushHandler)
       else:
         asyncSpawn wakuNode.publish(DefaultTopic, wakuMessage, conf.rlnRelay)
+
+proc getCustomTokens*() {.task(kind=no_rts, stoppable=false).} =
+  let tokens = status.getCustomTokens()
```
emizzle, line 638: Probably should check logged in state just like the other two tasks, as  it is required on the nim-status client side. We can keep it as it is, but there would be an inconsistent message to the end user, which may be confusing.

emizzle (commented):
Well done!! A few comments, nothing major, mostly preferences.

examples/chat/client/tasks.nim
siphiuel: Thanks, modified as suggested.

examples/chat/client/tasks.nim
siphiuel: Done.

examples/chat/tui/commands.nim
siphiuel: Fixed.

examples/chat/tui/commands.nim
siphiuel: Thanks, fixed!

examples/chat/tui/commands.nim
siphiuel: Agree. Do you mean we'd check that if we have just 4 (not 5 params), we'd then check if the fourth parameter begins with `0x`, and if so, try to parse it as hex, otherwise try to parse it as int?

examples/chat/tui/commands.nim
siphiuel: Looked at the body of StatusObject.getCustomTokens, and it seems to have try...except inside. It shouln't throw exceptions it seems. Should i modify it so that it does?

nim_status/client.nim
siphiuel: Removed casts, thanks for pointing this out.

nim_status/client.nim
siphiuel: Done.

examples/chat/tui/commands.nim
emizzle: Maybe we could try using `parseAddress` from `nim_status/conversions` with a try/catch, and if it fails, then we could consider it as invalid? Something like this:

```nim
  var addressParsed: Address
  try:
    addressParsed = address.parseAddress
  except Exception as e:
    # not valid, show validation error
```

siphiuel: `addCustomToken` in `tasks.nim` does the address parsing too and throws the exception if it fails, as per your suggestion below.

examples/chat/tui/commands.nim
emizzle: Actually that's a good point. We won't be able to get reliable exceptions from calls that are `asyncSpawn`ed, however I have seen it work, but while it may work sometimes, apparently not all exception are guaranteed to be catchable here. We may want to remove the other `try..except`s for `command.DeleteCustomToken` and `command.AddCustomToken` for consistency's sake then.

examples/chat/tui/commands.nim
```diff
@@ -27,6 +27,62 @@ const hashCharSet = {'#'}
...
+
+proc split*(T: type AddCustomToken, argsRaw: string): seq[string] =
+  argsRaw.split(" ")
+
+proc command*(self: ChatTUI, command: AddCustomToken) {.async, gcsafe,
+  nimcall.} =
+
+  try:
+    if command.address == "":
+      self.wprintFormatError(epochTime().int64,
```
emizzle, line 72: We believe we should try to move away from using `epochTime().int64` and instead use `getTime().toUnix` from the `std/times` lib.

emizzle (commented):
I also just noticed there are several instances of `epochTime().int64`. We have decided it's probably a good idea to move to `getTime().toUnix` for a non-localised int64 unix timestamp.

examples/chat/tui/commands.nim
siphiuel: Done.

examples/chat/tui/commands.nim
siphiuel: Fixed.

examples/chat/tui/commands.nim
emizzle: Right, but I think this is a separate issue - one is for validation (which helps to figure out which parameter the user was trying to specify),  and the other is for parsing the address to use in the nim-status client. One option would be to change the `AddCustomToken.address` type to `Address` and you won't need to worry about double-parsing.

examples/chat/tui/actions.nim
```diff
@@ -138,6 +138,79 @@ proc action*(self: ChatTUI, event: CreateAccountEvent) {.async, gcsafe,
...
+        let
+          name = token.name
+          symbol = token.symbol
+          address = token.address
+
+        self.printResult(fmt"{2.indent()}{name} ({symbol}): {address}", timestamp)
+        i = i + 1
+    else:
+      self.printResult(
+        "No custom tokens added. Add a custom token using `/addcustomtoken <address> <name> <symbol>`.",
```
emizzle, line 170: @michaelsbradleyjr has pointed out in one of my earlier PRs that the actions should not be command-dependent, and in that case the actions should not suggest commands.
```suggestion
        "No custom tokens added.",
```

examples/chat/tui/commands.nim
```diff
@@ -27,6 +27,59 @@ const hashCharSet = {'#'}
...
+      "custom token."),
+    CommandParameter(name: "name", description: "Name of " &
+      "the custom token."),
+    CommandParameter(name: "symbol", description: "Symbol of the " &
+      "custom token."),
+    CommandParameter(name: "color", description: "(Optional) Color (in hex) for " &
+      "custom token."),
+    CommandParameter(name: "decimals", description: "(Optional) Number of decimals to use for " &
+      "the custom token.")
+    ], description: "Creates a new custom token ")
```
emizzle, line 45:
```suggestion
    ], description: "Creates a new custom token.")
```

emizzle (approved):
Just some minor messaging recommendations. My apologies for not catching them on the first two rounds 😄

examples/chat/client/tasks.nim
```diff
@@ -729,3 +729,120 @@ proc publishWakuChat2*(message: string) {.task(kind=no_rts, stoppable=false).} =
...
+    trace "task sent event to host", event=eventNotLoggedInEnc, task
+    asyncSpawn chanSendToHost.send(eventNotLoggedInEnc.safe)
+    return
+
+  try:
+    let
+      deleteResult = status.deleteCustomToken(address.parseAddress) 
+
+    if deleteResult.isErr:
+      raise newException(Exception, deleteResult.error)
```
emizzle, line 831: Actually this is never guaranteed to be caught. It's better to send an error event to the task as you've already done above.

emizzle (changes requested):
Just caught another two things.

1. We may want to let the user delete a custom token by index rather than by address, similar to how we login.
2. There is an exception being raised inside of a task, which is not guaranteed to be caught anywhere. It's better to catch the exception (as you've done) and then send an error event back to the host.

nim_status/client.nim
```diff
@@ -362,6 +394,17 @@ proc toWalletAccount(account: accounts.Account): WalletAccount {.used.} =
...
   WalletAccount(address: account.address, name: name)
 
+proc getCustomTokens*(self: StatusObject): CustomTokensResult =
+  if not self.isLoggedIn:
+    return CustomTokensResult.err "Not logged in. Must be logged in to get " &
+      "custom tokens."
+  try:
+    let tokens = self.userDb.getCustomTokens()
+    return CustomTokensResult.ok tokens
+  except Exception as e:
```
arnetheduck, line 404: `Exception` should never be caught, it will cause issues with the distinction between `Defect` and `CatchableError` and doesn't work predictably - normally, more specific errors should be used instead, see - https://status-im.github.io/nim-style-guide/errors.exceptions.html

arnetheduck: In the case where unknown errors are filtering through, you can catch `CatchableError` instead, though we strive to avoid that

emizzle: Thanks for the tips @arnetheduck. After reviewing the style guide, there are quite a lot of changes we need to make as we are catching `Exception` in a lot of places. 

In this particular case, there are two branches of exceptions that we'd like to catch for this db call:
1. Any errors coming from the sql execution in the db (nim-sqlcipher lib)
2. An exception raised in the `userDb` proc (this needs to be updated to raise a custom error, but for this example let's call it `DbNotOpenError`)

What's the best way to proceed here? Would it be to catch both the `DbNotOpenError` and also to catch `CatchableError` which may come from any sql execution errors? Or should we try to modify nim-sqlcipher in a way that only raises custom errors?

examples/chat/tui/actions.nim
siphiuel: Fixed.

examples/chat/tui/commands.nim
siphiuel: Fixed.

nim_status/client.nim
arnetheduck: So there are two scenarios in general that are worth considering, I think - one is when dealing with user-supplied code, when you're making a "framework" - in that scenario, you can't anticipate what errors can happen but the nature of error handling tends to be limited to mitigating damage (closing sockets, releasing resources etc) and passing the exception on to someone more competent to handle it - this is not that common however (ie chronos and task_runner qualify but most other projects not) 

the other scenario is when you're creating abstraction layers - in this latter scenario, every non-custom error is an abstraction leak and therefore an architectural problem - ie when you're creating a "status client", the consumer of that client should not have to deal with "db error" from the db layer because there's nothing reasonable that they can do with this information beyond showing it to the user (which makes for terrible UX) - in particular, they are not in a position to import the concrete type of the error and reason about it, because that is an internal implementation detail of "the client". 

In such cases, the recommendation is that each layer handles _all_ exceptions originating from layers below it and either takes action (retries, repairs, reconnects, whatever) or reframes the error in terms of the "current" abstraction layer - "filesystem error" becomes "database error" becomes "cannot persist message" in the os, database and "message client" sections of the code respectively, each layer not only bundling functionality at a higher abstraction level, but also the possible errors that come out of there.

Finally, a layer such as `userDb` depends a little bit: if it adds its own "error hierarchy", one way to deal with it is to create a `UserDbError` in the general "Db" hierarchy of exceptions that has fields for the "inner" error that was raised from the `userDb` call, allowing consumers of these exceptions to catch `UserDbError` separately, _then_ drilling into it to find the "deeper" error if they care.

Important here is the design of any callback as well - callbacks that don't raise exceptions are easiest to reason about, but otherwise, you should document the callback itself with raises annotations (see https://status-im.github.io/nim-style-guide/language.proctypes.html)

> make as we are catching Exception in a lot of places.

I'm not sure if you're following it already, but the first and most important change is to start adding `{.push raises: [Defect].}` at the top of every module to enable _some_ error tracing in the language - the next step is to work on getting rid of `Exception` catches - part of this process is to _also_ sort out implicit Exception effects. There are more steps after that , but this is a start.

This is work that is best done early on in the project because inevitably it will need doing or the client will leak and crash and you won't be able to address it without rearchitecting the code completely to introduce the required error handling - as the other Nim projects we have have learned the hard way.

examples/chat/client/tasks.nim
siphiuel: Question - deleting by index then means we have to add ordering to the `tokens` DB table, right?

examples/chat/tui/commands.nim
siphiuel: Add address/decimals parsing to `commands.nim`. I've attempted to pass more specific types (`Address` and uint for address and decimals), but in that case `client/tasks.nim/AddCustomToken` compilation fails with
```
nim-status/vendor/nim-web3/web3/ethtypes.nim(231, 25) Error: ambiguous call; both typetraits.distinctBase(T: typedesc) [declared in nim-status/vendor/nimbus-build-system/vendor/Nim/lib/pure/typetraits.nim(70, 6)] and sugar.distinctBase(T: typedesc) [declared in nim-status/vendor/nimbus-build-system/vendor/Nim/lib/pure/sugar.nim(164, 10)] match for: (FieldType)
```
Have no idea what this is about, so left it untyped for now.

emizzle: Maybe try instead of importing all of the sugar module, import only `=>`, ie:
```nim
import std/sugar #<== how it currently is
# to
from std/sugar import `=>`
```

examples/chat/client/tasks.nim
emizzle: Hmm, yes good point. We could order by symbol maybe? What do you think?

siphiuel: Added ordering by symbol. Did it yesterday actually (by address), but forgot to click the `Comment` button.

siphiuel: Also - modified DeleteCustomToken to accept `index` parameter.

examples/chat/tui/commands.nim
siphiuel: Thanks for the tip, this helped!

siphiuel:
@emizzle replaced Exception `raise`s with error events.

michaelsbradleyjr (approved):
I read through it all again today — looks good!

I did notice that some `except Exception` (added in this PR vs. already in place) still remain, but given that @emizzle is working on a refactor that will address all of them, I think that's okay.

On the other hand, a quick touch up could be to change any new (per this PR) `except Exception` to `except CatchableError`.

nim_status/client.nim
emizzle: Thanks so much for the writeup @arnetheduck. @michaelsbradleyjr and I have read through it thoroughly a few times and it has provided quite a lot of insight. We discussed it further and it resulted in a small [exception handling proposal](https://discord.com/channels/624205794384281629/720285913858703391/869740387089125376), which is based on our limited understanding. The first part of this is in draft (https://github.com/status-im/nim-status/pull/246) and we would love to get your feedback on it if possible 🙏

PR #240, opened 2021-07-26
feat: delete wallet account

emizzle:
Closes #225.

Deletes wallet account by wallet index.

Add account db tests for additional account procs: getWalletAccount, getWalletAccounts, and deleteWalletAccount

Add account generator tests for additional generator procs: deleteKeyFile

PR #242, opened 2021-07-27
feat: en/decode Status payloads when sending/receiving messages (first steps)

### LINKED ISSUE #217: encode / decode message payload when sending/receiving messages
currently messages are sent plainly, goal is to:

- when a message is sent, it should encoded with protobuffers with the payload structure of https://specs.status.im/spec/6#message
- when a message is received it should decode the payload and display the message

michaelsbradleyjr:
Copy protobuf specs from status-go, make necessary adjustments for use via nim-protobuf-serialization. Many thanks to @richard-ramos for his help with this aspect and others!

Implement "first steps" towards encoding and decoding Status payloads when sending and receiving messages.

That consists of using an incomplete `ProtocolMessage(ApplicationMetadataMessage(ChatMessage)))` wrapping of messages, assuming all messages currently being sent correspond to public chats. It's incomplete because there are many fields for which values are not being derived, and there is no encryption performed.

A series of future pull requests will fill in the missing pieces and expand our usage of the protobuf specs, i.e. allow nim-status to send and receive more kinds of Status messages.

Note that the waku v2 content topics are not yet using a hash of the user-supplied topic name; that will also be addressed in a future pull request.

It's also important to note that a refactor is pending re: moving the import and usage of nim-waku inside the `nim_status` library, as well refactorings re: exception handling, imports and exports, and other aspects that have been commented upon in other recent pull requests. This commit builds on the work done to date in the same way as before, but there is already awareness of what should be changed and can be improved.

Ensure that joined waku v2 content topics (whether joined in the TUI or specified on the command-line) conform to the recommendations in [23/WAKU2-TOPICS](https://rfc.vac.dev/spec/23/).

Unrelated to the main focus of this commit: use a fixed-length array instead of a sequence for collecting bytes of user input (keyboard and mouse) in the `input` worker thread. It's more efficient to do it that way: less allocation and garbage collection.

---

Closes #217.


examples/chat/client/tasks.nim
```diff
@@ -81,25 +92,76 @@ proc statusContext*(arg: ContextArg) {.async, gcsafe, nimcall,
   resetContext()
 
 proc new(T: type UserMessageEvent, wakuMessage: WakuMessage): T =
-  let topic = wakuMessage.contentTopic
+  let
+    topic = wakuMessage.contentTopic
+    topicSplit = topic.split('/')
```
emizzle, line 97: From the perspective of someone who hasn't been coding this feature for some time, it would make it easier to digest if the `contentTopic` was parsed in to an object, and from there, each constituent part could be identified with a name. Something like this:
```nim
import
  std/[sequtils, strutils]

type
  Encoding* {.pure.} = enum
    Protobuffer* = "proto"
    RLP* = "rlp"
    
  ContentTopic* = ref object
    appName*: string
    appVersion*: int
    contentTopicName*: string
    encoding*: Encoding

  ParseError* = object of ExampleClientError # Maybe we should create this error type

proc parseContentTopic(contentTopic: string): ContentTopic {.raises: [Defect, ParseError].} =
  let topicSplit = contentTopic.split('/').keepIf(t => t != "") # remove empties
  
  if topicSplit.len != 4: raise ParseError fmt"'{contentTopic}' has an insufficient number of parts"
  
  var version: int
  try:
     version = topicSplit[1].parseInt
  except ValueError as e:
    raise ParseError fmt"Could not parse '{topicSplit[1]}' as a version integer"
  
  var encoding: Encoding
  try:
    encoding = parseEnum[Encoding](topicSplit[3])
  except ValueError as e:
    raise ParseError fmt"Could not parse '{topicSplit[3]}' as an encoding"

  return ContentTopic(
    appName: topicSplit[0]
    appVersion: version,
    contentTopicName: topicSplit[2],
    encoding: encoding
  )
```
And then the following would change:
```suggestion
var topic: ContentTopic
try:
  topic = wakuMessage.contentTopic.parseContentTopic
except ParseError as e:
  # do something with the error
  error "Failed to parse content topic, contentTopic=wakuMessage.contentTopic
```
Which could then be used like:
```nim
case fmt"{topic.appName}/{topic.appVersion}":
  of toyChat2App:
    if topic.encoding == Encoding.Protobuffer:
      # ...
```

examples/chat/client/tasks.nim
```diff
@@ -710,22 +777,65 @@ proc lightpushHandler(response: PushResponse) {.gcsafe.} =
...
+    for topic in contentTopics:
+      var
+        payload: seq[byte]
+        unsupported = false
+
+      let topicSplit = topic.split('/')
+
+      case fmt"{topicSplit[1]}/{topicSplit[2]}":
+        of toyChat2App:
+          if topicSplit[4] == "proto":
```
emizzle, line 802: Same as my comment above regarding parsing the `ContentTopic`.

examples/chat/tui/actions.nim
```diff
@@ -309,7 +309,7 @@ proc action*(self: ChatTUI, event: UserMessageEvent) {.async, gcsafe, nimcall.}
 
     # if event.topic is not a properly formatted waku v2 content topic then the
     # whole string will be passed to printMessage
-    if topicSplit.len == 5:
+    if topicSplit.len == 5 and topicSplit[0] == "":
       # for "/toy-chat/2/example/proto", topic would be "example"
       topic = topicSplit[3]
```
emizzle, line 314: Even here, I think we could parse a `ContentTopic` object, or perhaps `UserMessageEvent.topic` could be of type `ContentTopic` and it would be parsed in the task.

nim_status/accounts/accounts.nim
```diff
@@ -74,6 +74,23 @@ proc getAccounts*(db: DbConn): seq[Account] =
...
+                              {tblAccounts.`type`.columnName},
+                              {tblAccounts.storage.columnName},
+                              {tblAccounts.path.columnName},
+                              {tblAccounts.publicKey.columnName},
+                              {tblAccounts.name.columnName},
+                              {tblAccounts.color.columnName},
+                              {tblAccounts.createdAt.columnName},
+                              {tblAccounts.updatedAt.columnName}
+                    FROM      {tblAccounts.tableName}
+                    WHERE     {tblAccounts.chat.columnName} = TRUE"""
```
emizzle, line 91: We are using `0`/`1` instead of `FALSE`/`TRUE` in `getWalletAccounts` (and a couple others in #240). I have no strong preference either way, but maybe we should keep it consistent?

nim_status/protocol.nim
```diff
@@ -0,0 +1,6 @@
+import # nim-status libs
+  ./protocol/application_metadata_message,
+  ./protocol/chat_message,
+  ./protocol/encryption/protocol_message
```
emizzle, line 4:
```suggestion
  ./protocol/[application_metadata_message, chat_message,
              encryption/protocol_message]
```

emizzle: No strong preference either way here.

nim_status/protocol/application_metadata_message.nim
```diff
@@ -0,0 +1,14 @@
...
+
+# nim-status protobuf spec
+# automatically exports `type ApplicationMetadataMessage`
+import_proto3 "protobuf/application_metadata_message.proto"
+
+proc decode*(T: type ApplicationMetadataMessage, input: seq[byte]): T =
+  Protobuf.decode(input, T)
+
+proc encode*(input: ApplicationMetadataMessage): seq[byte] =
+  Protobuf.encode(input)
```
emizzle, line 14: Are these really necessary? Wouldn't this be sort of "ungenericizing" (or whatever the word is) the already generic `Protobuf.encode`/`Protobuf.decode`? It would mean that we would need to include these procs for each protobuf object introduced which can be more difficult to maintain.

nim_status/protocol/chat_message.nim
```diff
@@ -0,0 +1,14 @@
...
+export protobuf_serialization
+
+# nim-status protobuf spec
+# automatically exports `type ChatMessage`
+import_proto3 "protobuf/chat_message.proto"
+
+proc decode*(T: type ChatMessage, input: seq[byte]): T =
+  Protobuf.decode(input, T)
+
+proc encode*(input: ChatMessage): seq[byte] =
```
emizzle, line 13: Same comment.

nim_status/protocol/encryption/protocol_message.nim
```diff
@@ -0,0 +1,14 @@
...
+
+# nim-status protobuf spec
+# automatically exports `type ProtocolMessage`
+import_proto3 "protobuf/protocol_message.proto"
+
+proc decode*(T: type ProtocolMessage, input: seq[byte]): T =
+  Protobuf.decode(input, T)
+
+proc encode*(input: ProtocolMessage): seq[byte] =
+  Protobuf.encode(input)
```
emizzle, line 14: Same comment.

emizzle (approved):
Overall, really fantastic work and a mammoth effort!! 

I'd say regardless of my comments, nothing is, in my opinion, imperative to change (hence the approval), however I do have some suggestions which would make it more readable for other developers.

examples/chat/client/tasks.nim
michaelsbradleyjr: Thank you for writing this up! Excellent recommendations.

One of the tricky things is that content topic formatting is only a recommendation in the waku v2 specs. To my understanding, some apps using waku v2 (e.g. WalletConnect) use a different scheme for content topic strings.

On the other hand, as is already the case, the example client can be opinionated and enforce the recommended scheme. In which case having types and errors and a parser proc makes a a lot of sense — less fragile and more DRY, for sure.

Once we pull waku inside the `nim_status` library, the types, etc. you suggested should also be implemented within `nim_status` and offered as a re-export of `nim_status/client.nim`. The the example client's commands, tasks, etc. can make use of that facility.

If you're okay with it, let's hold off on implementing those helpers in this PR and do it as part of the refactor to pull waku inside `nim_status` (or a PR that follows on that refactor).

nim_status/accounts/accounts.nim
michaelsbradleyjr: SQLite (and so SQLCipher) auto-translates `FALSE / TRUE` in queries to `0 / 1` and on the nim side we have the field as `Option[bool]`, so I think it's more natural to use `FALSE / TRUE`. At the same time, consistency is important, so we should decide which way we want to do it and then change all the queries to use the same style.

nim_status/protocol.nim
michaelsbradleyjr: Changed.

nim_status/protocol/application_metadata_message.nim
michaelsbradleyjr: It allows for this:
```nim
let amm = ApplicationMetadataMessage()

let
  enc = amm.encode
  dec = ApplicationMetadataMessage.decode(enc)
```
Without it:
```nim
let amm = ApplicationMetadataMessage()

let
  enc = Protobuf.encode(amm)
  dec = Protobuf.decode(enc, ApplicationMetadataMessage)
```
I prefer the fist style. But:
> It would mean that we would need to include these procs for each protobuf object...

It's true we'd need them for each protobuf, but once we have them in a corresponding `.nim` module for each `.proto` spec, they are unlikely to change over time, even if we update the specs, i.e. copy over updated specs from status-go. So the effort to create them and re-export them from `nim_status/protocol.nim` seems like a mostly fixed development cost.

Issue #239, opened 2021-07-26
Remove all instances of `except Exception as e:` and review exception handling


emizzle:
Based on Jacek's comment in https://github.com/status-im/nim-status/pull/237#discussion_r676405120.

emizzle:
Proc `userDb()` in `nim-status/client.nim` needs a custom error as per the style guide: https://status-im.github.io/nim-style-guide/errors.exceptions.html

PR #246, opened 2021-07-29
feat: first steps to enable proper exception handling

emizzle:
Related to #239.

This PR represents the first steps to tackle [@arnetheduck's suggestions](https://github.com/status-im/nim-status/pull/237#discussion_r676531575) regarding proper error handling in the nim-status library.

This PR is meant to cover #1 in the [exception handling proposal](https://discord.com/channels/624205794384281629/720285913858703391/869740387089125376) (PR #1 (easy wins): Swap all `except Exception as e` for `except CatchableError as e` and add `{.push raises: [Defect].}` at the top of all modules). PR #2 (exception hierarchy) will be covered in a future PR.

There are a few questions I have that I would love to understand more of if possible, and apply the learnings to this PR:
1. In quite a lot of procs, the compiler is forcing a listing of `{.raises:[Defect, Exception].}`. There is something that is not correct here, right? We should never raise `Exception` directly, right?
~~2. Must we always include `Defect` in each `{.raises.}` pragma? It seems that the compiler provides hints that this is not always necessary: `Hint: 'Defect' is declared but not used [XDeclaredButNotUsed]`~~
3. There are some procs that we know won't raise a particular type of exception, how can we avoid listing it in `raises`? Consider the following:

    ```jsx
    proc findAccount(self: Generator, accountId: UUID): AccountResult =
      let id = $accountId
      if not self.accounts.hasKey(id):
        return AccountResult.err "Account doesn't exist"

      AccountResult.ok(self.accounts[id])
    ```

    With `{.push raises: [Defect].}` at the top of the module, the compiler will complain about this proc as follows:

    ```jsx
    Error: can raise an unlisted exception: ref KeyError
    ```

    But we know it should never happen due to the key check. Is there a way to explicitly tell the compiler that this should not be raised so that it doesn't need to be included in the `{.raises: [,,].}` list?

4. When should we use `ref MyError` vs just `MyError`?


michaelsbradleyjr:
I just rebased this PR branch against current `master`.

emizzle:
As discussed with @michaelsbradleyjr, we are going to merge this PR as is with the incorrect raising of `Exception`s (`{.raises: [Exception].}`. The next PR to come will work on reducing the number of  places this occurs, minimising it where possible, as well as attempting to have no errors raised where possible in the public-facing API, opting to abstracting them out with `Result[T, string]`.

arnetheduck:
Oops, looks like I missed some of the questions in the description here!

> In quite a lot of procs, the compiler is forcing a listing of {.raises:[Defect, Exception].}. There is something that is not correct here, right? We should never raise Exception directly, right?

`Exception` is often deduced by the compiler as a "worst-case" - for example forward declarations and proc types are assumed to "raise anything" as far as the exception tracking mechanism goes - this has a tendency to spread into modules - we sometimes use the following to contain such diseases, when we can't fix the original code:

```nim
try:
   ...
except Exception as e:
  if e of Defect:
   raise (ref Defect)(e)
  elif e of CatchableError:
   raise (ref CE)(e)
  else:
    raise (ref CatchableError)(msg: e.msg, parent: e)
```

> With {.push raises: [Defect].} at the top of the module, the compiler will complain about this proc as follows:

```nim
try:
   a[b]
except KeyError:
  raiseAssert "Should never happen"
```

this turns the unexpected exception into a Defect/assert

> When should we use ref MyError vs just MyError?

Mostly when constructing exceptions: `raise (ref XxxError)(msg: "...")` - otherwise it's rarely needed, the compiler does clever dark magic when matching exception types

PR #247, opened 2021-07-29
refactor: nim_status -> status

michaelsbradleyjr:
This is a redo of #243.

---

The name of the repo is ***not*** being changed. But instead of doing:

```nim
import nim_status
```

You would do:

```nim
import status
```

This is in line with nim-chronos (`import chronos`), nim-waku (`import waku`), and many other Nim libs.


michaelsbradleyjr:
I intend to merge if/when the GHA workflows complete successfully.

PR #249, opened 2021-07-30
Fetch token prices from cryptocompare.io

### LINKED ISSUE #238: Get Fiat price for a Token

siphiuel:
Fixes #238.

**Debugging note**: it is important to properly initialise `Table` objects. See https://github.com/status-im/nim-status/blob/307a321e2530d0de904b0b6dc18793e95fe56682/status/api/common.nim#L54.

Otherwise some operations on those objects might crash with `SIGSEGV`s, which is quite difficult to debug (e.g., writing to a `var` works fine, but reading from it causes a crash). Kudos to @michaelsbradleyjr for helping to catch that one.

### DIFF
--- a/examples/client/client/tasks.nim
+++ b/examples/client/client/tasks.nim
@@ -51,6 +51,8 @@ var
   subscribed {.threadvar.}: bool
   status {.threadvar.}: StatusObject
   statusState {.threadvar.}: StatusState
+  updatePricesTimeout {.threadvar.}: int
+  updatingPrices {.threadvar.}: bool
   wakuNode {.threadvar.}: WakuNode
   wakuState {.threadvar.}: WakuState
 
@@ -85,6 +87,9 @@ proc statusContext*(arg: ContextArg) {.async, gcsafe, nimcall,
   # previous comment
   nodekeyGenerated = false
 
+  updatePricesTimeout = 60000 # 60 seconds by default
+  updatingPrices = false
+
   status = StatusObject.new(conf.dataDir).expect(
     "StatusObject init should never fail")
   # threadvar `statusState` is currently out of scope re: "resetting the
@@ -94,6 +99,18 @@ proc statusContext*(arg: ContextArg) {.async, gcsafe, nimcall,
 
   # re/set threadvars that don't persist across waku dis/connect
   resetContext()
+
+proc updatePrices() {.async.} =
+  updatingPrices = true
+
+  while statusState == StatusState.loggedin:
+    let res = await status.updatePrices()
+    if res.isErr:
+      # TODO: send error to the TUI for display (possibly in the status bar?)
+      error "failed to update prices", error = $res.error
+    await sleepAsync(updatePricesTimeout)
+
+  updatingPrices = false
 
 proc new(T: type UserMessageEvent, wakuMessage: WakuMessage): T =
   let
@@ -611,6 +628,8 @@ proc login*(account: int, password: string) {.
 
     statusState = StatusState.loggedin
 
+    if not updatingPrices: asyncSpawn updatePrices()
+
     event = LoginEvent(account: publicAccount, error: "", loggedin: true)
     eventEnc = event.encode
 
@@ -1123,6 +1142,53 @@ proc callRpc*(rpcMethod: string, params: JsonNode) {.task(kind=no_rts, stoppable
     trace "task sent event with error to host", event=eventEnc, task
     asyncSpawn chanSendToHost.send(eventEnc.safe)
 
+proc getPrice*(tokenSymbol: string, fiatCurrency: string) {.task(kind=no_rts, stoppable=false).} =
+  let timestamp = getTime().toUnix
+
+  if statusState != StatusState.loggedin:
+    let
+      eventNotLoggedIn = GetPriceEvent(error: "Not logged in, " &
+        "cannot get price for a custom token.",
+        timestamp: timestamp)
+      eventNotLoggedInEnc = eventNotLoggedIn.encode
+      task = taskArg.taskName
+
+    trace "task sent event to host", event=eventNotLoggedInEnc, task
+    asyncSpawn chanSendToHost.send(eventNotLoggedInEnc.safe)
+    return
+
+  let price = status.getPrice(tokenSymbol, fiatCurrency)
+  if price.isErr:
+    let
+      event = GetPriceEvent(error: $price.error)
+      eventEnc = event.encode
+      task = taskArg.taskName
+
+    trace "task sent errored event to host", event=eventEnc, task
+    asyncSpawn chanSendToHost.send(eventEnc.safe)
+    return
+
+  let
+    event = GetPriceEvent(symbol: tokenSymbol, currency: fiatCurrency,
+      price: price.get, timestamp: getTime().toUnix)
+    eventEnc = event.encode
+    task = taskArg.taskName
+
+  trace "task sent event to host", event=eventEnc, task
+  asyncSpawn chanSendToHost.send(eventEnc.safe)
+
+proc setPriceTimeout*(priceTimeout: int) {.task(kind=no_rts, stoppable=false).} =
+  let timestamp = getTime().toUnix
+
+  updatePricesTimeout = priceTimeout * 1000
+
+  let
+    event = SetPriceTimeoutEvent(timeout: priceTimeout, timestamp: getTime().toUnix)
+    eventEnc = event.encode
+    task = taskArg.taskName
+
+  trace "task sent event to host", event=eventEnc, task
+  asyncSpawn chanSendToHost.send(eventEnc.safe)
 
 proc sendTransaction*(fromAddress: eth_common.EthAddress,
   transaction: eth_common.Transaction, password: string)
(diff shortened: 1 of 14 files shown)


examples/client/client/tasks.nim
```diff
@@ -92,6 +93,17 @@ proc statusContext*(arg: ContextArg) {.async, gcsafe, nimcall,
   # re/set threadvars that don't persist across waku dis/connect
   resetContext()
 
+proc updatePrices() {.async.} =
+  while statusState == StatusState.loggedin:
+    await sleepAsync(5000)
```
michaelsbradleyjr, line 98: I wonder if 5 seconds is too frequent. Maybe it should be 60 seconds?

Maybe the `sleepAsync` should be moved to the bottom of the loop so that as soon as `status.login` succeeds, the first update will run right away.

I don't think it's too important now, but in the future we could add a command that allows the frequency to be set by the user. The corresponding task in `client/tasks.nim` would simply update the value of a threadvar referenced by the `updatePrices` looper proc.

examples/client/client/tasks.nim
```diff
@@ -1136,3 +1150,41 @@ proc callRpc*(rpcMethod: string, params: JsonNode) {.task(kind=no_rts, stoppable
...
+        "cannot get price for a custom token.",
+        timestamp: timestamp)
+      eventNotLoggedInEnc = eventNotLoggedIn.encode
+      task = taskArg.taskName
+
+    trace "task sent event to host", event=eventNotLoggedInEnc, task
+    asyncSpawn chanSendToHost.send(eventNotLoggedInEnc.safe)
+    return
+
+
```
michaelsbradleyjr, line 1169: Can get rid of the extra blank line.

examples/client/tui/common.nim
```diff
@@ -178,6 +183,7 @@ const
     "create": "createaccount",
     "delete": "deletewalletaccount",
     "deletetoken": "deletecustomtoken",
+    "getprice": "getprice",
```
michaelsbradleyjr, line 186: No need to add to the aliases table if there's no alias for `getprice`.

examples/client/tui/common.nim
```diff
@@ -209,6 +215,7 @@ const
     "deletewalletaccount": @["delete"],
     "getassets": @["assets"],
     "getcustomtokens": @["gettokens"],
+    "getprice": @["getprice"],
```
michaelsbradleyjr, line 218: No need to add to the aliased table if there's no alias for `getprice`.

michaelsbradleyjr (changes requested):
Nice work! I'm glad we were able to finally get to the bottom of the runtime crashes.

examples/client/client/tasks.nim
```diff
@@ -92,6 +93,17 @@ proc statusContext*(arg: ContextArg) {.async, gcsafe, nimcall,
   # re/set threadvars that don't persist across waku dis/connect
   resetContext()
 
+proc updatePrices() {.async.} =
+  while statusState == StatusState.loggedin:
+    await sleepAsync(5000)
+    if statusState == StatusState.loggedout:
```
michaelsbradleyjr, line 99:
```suggestion
    if statusState != StatusState.loggedin:
```

Should help avoid ambiguity re: `StatusState.loggingin` and `StatusState.loggingout`.

michaelsbradleyjr (changes requested):

examples/client/client/tasks.nim
siphiuel: Done.

examples/client/tui/common.nim
siphiuel: Done.

examples/client/tui/common.nim
siphiuel: Done.

examples/client/client/tasks.nim
siphiuel: Please see above comment.

examples/client/client/tasks.nim
```diff
@@ -610,6 +622,8 @@ proc login*(account: int, password: string) {.
 
     statusState = StatusState.loggedin
 
+    asyncSpawn updatePrices()
```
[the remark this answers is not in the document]
siphiuel: Thanks for a great observation! 

What if we simplify the `updatePrices` proc so that it looks like this:
```
proc updatePrices() {.async.} =
  while statusState == StatusState.loggedin:
    try:
      await status.updatePrices()
    except CatchableError as e:
      trace "updatePrices error:", err=e.msg
    await sleepAsync(updatePricesTimeout)
```
So it will run in the background, periodically trying to re-fetch prices if user is logged in currently. I think it's fine that the updater is started during the first login, and i've moved the `sleepAsync` to the end of the loop body, as you suggested. The prices are not account-dependent anyway.

I think i've accidentally clicked `Resolve conversation` on one of your comments, but it pertained to `updatePrices` anyway i think.

siphiuel:
@michaelsbradleyjr added a `/setpricetimeout` command, as per your suggestion i've accidentally "resolved".

examples/client/client/tasks.nim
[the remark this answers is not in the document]
michaelsbradleyjr: I reviewed all the changes made and they look good. Thanks for adding the `/setpricetimeout` command, very nice!

However, there's still the problem of multiple updater loops being spawned.

Take a look at these changes: https://github.com/status-im/nim-status/compare/c26f0321b24aa5d5f4f010851054b4a7984c616c...feat/token-prices-2.

In that branch I've rebased on top of my `fix/help` branch for PR #277, made same changes for commands you introduced, and then I added some changes to `examples/client/client/tasks.nim` to demonstrate the problem and proposed fix.

The code for the proposed fix is commented out.

Note that I also removed the `loggedin` check in `proc setPriceTimeout` in `client/tasks.nim` since there's no good reason to have it.

To see the problem manifest:
* Build and run the client, then `/setpricetimeout 20`, `/login ...`, wait until login is successful, and then `/logout` and `/login ...` several times before 20 seconds is up.
* Watch the log: in intervals of 20 seconds you'll see that multiple loop "instances" are running.

If you uncomment the proposed fix (lines involving threadvar `updatingPrices`) and repeat the experiment, there will only be one instance of the loop running at any given time when `loggedin`.

---

Unrelated to the above, I see that an exception is being raised every time the loop runs; in the log it looks like:

```
TRC 2021-08-17 13:56:14.750-05:00 Exception when parsing price API response  topics="client" tid=16054077 file=tokens.nim:139 err="key not found: RAW"
```

michaelsbradleyjr (changes requested):
My apologies for another requested change, don't mean to drag this out, but re: bugfix #277, which just landed on `master`, procs added in `tui/commands.nim` should use `aliased.getOrDefault(command)` instead of `aliased[command]`.

Those changes are already made in https://github.com/status-im/nim-status/compare/63dae07...feat/token-prices-2, so can be easily applied.

siphiuel:
@michaelsbradleyjr thank you! Applied the requested fixes.

examples/client/client/tasks.nim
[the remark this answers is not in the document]
siphiuel: Regarding the exception - it's a normal flow in case when the URL is malformed (e.g. when we have no tokens to fetch prices for). Maybe this can be handled by a conditional in another part of code.

michaelsbradleyjr (approved):

michaelsbradleyjr:
After approving the PR, I had occasion to look again at `private/tokens.nim` and realized a couple of refactors were desirable:
* `updatePrices` proc in both `api/tokens` and `private/tokens` should use `Result`.
* There was a stray `logScope` in `private/tokens` that seems to have been copied over from the example client.

I've pushed a commit that addresses both things. It's probably not in ideal shape as I'm still getting up to speed with the best way to work with `Result` per @emizzle's PR #269. I'd like to get his feedback on my changes before merging this PR, but we should be able to merge it before the end of my day today.

emizzle:
After rebasing this against master, I added a couple of changes in 516be451. Firstly, note that I messed up the rebase in tasks.nim 516be451, but fixed it in 0e85f3a6.

After giving this some thought, it seemed more apparent that the HTTP API requests for crypto compare and open sea were similar, and could share a Result type of `HttpFetchResult*[T] = Result[T, HttpFetchError]`. If we add in more external HTTP API calls, we could utilise this type again. This type (`HttpFetchResult`) is at nim-status private layer, so I moved the OpenSea fetches to a private module. 

The motivation for this change was to attempt at adhering to move from `general -> specific` when going from `private -> public` (low to high layers). For example, there are lots of modules at the private lower layer that use `DbResult[T]`, and then in the higher public API layer, they use a more specific Result type that pertains to that particular API module (ie `SettingsResult` in the settings API module). Any API module could consume several lower-layer Result types and "translate" them to its own specific Result type for its module. This was the motivation for moving external HTTP API fetching to the same result type in the lower private layer, as consumers in the upper API layer end up translating the errors for their consumers.

I would like to extrapolate this a bit further in regards to  `private/callrpc.nim` and `api/provider.nim`, as I think the layers there are a bit convoluted, but will handle this at some point in the future.

examples/client/client/tasks.nim
```diff
@@ -95,6 +100,16 @@ proc statusContext*(arg: ContextArg) {.async, gcsafe, nimcall,
   # re/set threadvars that don't persist across waku dis/connect
   resetContext()
 
+proc updatePrices() {.async.} =
+  updatingPrices = true
+
+  while statusState == StatusState.loggedin:
+    let res = await status.updatePrices()
+    if res.isErr: error "failed to update prices", error = $res.error
```
michaelsbradleyjr, line 108: If/when it fails, we should probably send an event to the TUI so the user can be made aware of it.

emizzle: I don't have a strong opinion either way, but `updatePrices` is executed as a background operation, so it may seem odd that the user sees a message for an operation they did not start. However, it could be a message that is displayed in the "status bar" in the future. I'll add it in.

emizzle: Actually, it's not that simple to add in because this is not within the context of a task. Will add a comment and leave it for now to get this merged.

PR #251, opened 2021-08-01
feat(client): basic support for backspace

michaelsbradleyjr:
Note: on mac keyboards the delete key is backspace and on mac laptop keyboards fn+delete is delete ("forward delete"). Some mac keyboards have a separate key, also named delete, for performing forward deletion.

---

After I've tested locally on Linux and Windows, I'll take this PR out of draft.

Tested on:
- [x] macOS
- [x] Linux
- [x] Windows

### DIFF
--- a/examples/client/tui/actions.nim
+++ b/examples/client/tui/actions.nim
@@ -27,6 +27,8 @@ logScope:
 # at least 60 FPS in the TUI.
 
 const
+  KEY_BACKSPACE = "KEY_BACKSPACE"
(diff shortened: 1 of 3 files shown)


siphiuel (approved):
Wow, this is great! Will make the TUI so much more comfortable.

emizzle (approved):
Works really well, well done and THANK YOUUUU!! 

In the future, it would be great to support option+backspace to delete a word, but obviously out of scope for this PR.

PR #254, opened 2021-08-02
chore: status/client -> status/api, modularise api

emizzle:
Closes: #253 

Change the name from `status/client.nim` to `status/api.nim`.

Break up `api.nim` in to modules, and put the modules in to the `status/api` directory.

Update the tests (more work is needed, issue created: https://github.com/status-im/nim-status/issues/252).


status/settings.nim
```diff
@@ -92,13 +92,9 @@ proc getSettings*(db: DbConn): Settings =
     raise newException(ValueError, "No record found for settings")
   settings.get
 
-proc saveSetting*(db: DbConn, setting: string, value: auto)
-  {.deprecated("Please use saveSetting(SettingsCol) instead").} =
+proc saveSetting*(db: DbConn, setting: string | SettingsCol, value: auto) =
```
siphiuel, line 95: Cool, i didn't know Nim supports alternative types for proc args! Can you please send a link to the Nim manual where it's described? I tried finding it myself but failed.

siphiuel (approved):

status/settings.nim
michaelsbradleyjr: I think it's allowed because `SettingsCol` type is a pure enum where all the enum values correspond to type `string`.

In which case, is there a reason to prefer `string | SettingsCol` to `SettingsCol`?

examples/client/client/events.nim
```diff
@@ -5,7 +5,7 @@ import # client libs
   ./common
 
 import # status libs
-  ../../../status/[client, tokens]
+  ../../../status/[api, tokens]
```
michaelsbradleyjr, line 8: This is an example where there's a mix of importing the api and non-api modules. See my main comment above; again, I understand this PR is a first step in the reorg/refactor.

status/api/accounts.nim
```diff
@@ -0,0 +1,187 @@
...
+  ./common, ../conversions, ../extkeys/[paths, types], ../identicon, ../settings
+
+export
+  accounts, alias, common, conversions, generator, identicon, paths,
+  public_accounts, secp256k1, settings, types, uuid
+
+type
+  PublicAccountResult* = Result[PublicAccount, string]
+
+proc storeDerivedAccounts(self: StatusObject, id: UUID, keyUid: string,
```
michaelsbradleyjr, line 20: In our api modules, would it make sense to attempt to only include procs that are public/exported?

Thinking about it, I see how there can arise some tricky situations re: circular dependencies. Also, at times it wouldn't be worth the effort to refactor a non-exported helper into a non-api module (from which it would be exported, but not re-exported by the api module).

In a lot of cases, maybe it would be a worthwhile exercise to at least consider whether a non-exported proc could be moved and what kinds of entanglements are involved. Or maybe not... needs more consideration.

michaelsbradleyjr (approved):
This looks good.

My main question:

Eventually, would we want that a consumer of nim-status, such as our example client, could simply `import status`, i.e. all the things it will need are re-exported from the root `status.nim`?

If so, then the root `status.nim` could become:
```nim
import # status libs
  ./status/api

export api
```

I know we've discussed it before, and this PR seems more like a first step, but I wanted to get your thoughts again as we move forward. I also recall this involves a trade-off re: longer compile times.

I would expect that in unit tests there could be tests for all the things re-exported by `status/api.nim` through the root `status.nim`.

However, in some cases it could be important (and/or more convenient) in the unit tests to import non-api modules (i.e. modules other than root `status.im`, `status/api.nim`, `status/api/*.nim`) and test them directly rather than trying to test them through the public api.

status/settings.nim
emizzle: > Cool, i didn't know Nim supports alternative types for proc args! Can you please send a link to the Nim manual where it's described? I tried finding it myself but failed.

The closest explanation in the manual I could find is in the [Generics Type Classes](https://nim-lang.org/docs/manual.html#generics-type-classes) and the [Implicit generics](https://nim-lang.org/docs/manual.html#generics-implicit-generics) sections:
> "Whilst the syntax of type classes appears to resemble that of ADTs/algebraic data types in ML-like languages, it should be understood that type classes are static constraints to be enforced at type instantiations. Type classes are not really types in themselves but are instead a system of providing generic "checks" that ultimately resolve to some singular type. Type classes do not allow for runtime type dynamism, unlike object variants or methods.
> A type class can be used directly as the parameter's type. Procedures utilizing type classes in such a manner are considered to be implicitly generic. They will be instantiated once for each unique combination of param types used within the program.
> By default, during overload resolution, each named type class will bind to exactly one concrete type. We call such type classes bind once types."

emizzle: > In which case, is there a reason to prefer `string | SettingsCol` over `SettingsCol`?

There was originally a deprecated proc `saveSetting` that accepted `setting: string`. I later added the second `saveSetting` proc that used `setting: SettingsCol` which called the original deprecated proc. Considering the deprecated proc was still called, it was still causing a compiler warning to be issued, and since the implementation of the deprecated proc simply used a `$` operation (via `fmt"{setting}"`), changing the parameter type to accept both `string` and `SettingCol` worked without changing the implementation, got rid of the deprecation, and still supported `string` as an alternative parameter type. Retaining support for `string` param type assumes that there are some external consumers of `nim-status` that are using this proc, however I highly doubt there are any in the wild, so **I've now changed this to simply remove `string` support, and only support `SettingsCol`**.

emizzle:
> I know we've discussed it before, and this PR seems more like a first step, but I wanted to get your thoughts again as we move forward. I also recall this involves a trade-off re: longer compile times.

I definitely agree, as we discussed, we desire consumers of `nim-status` to simply `import status` which will have access to all public-facing api modules re-exported from the root `status.nim`, which effectively will include:
- `status/api.nim`
- `status/api/*.nim` (re-exported from `status/api.nim`)

As was pointed out to us, this would induce longer compile times as a trade off to reduce the overall exposed API surface area.

> I would expect that in unit tests there could be tests for all the things re-exported by status/api.nim through the root status.nim.

>However, in some cases it could be important (and/or more convenient) in the unit tests to import non-api modules (i.e. modules other than root status.im, status/api.nim, status/api/*.nim) and test them directly rather than trying to test them through the public api.

I agree wholeheartedly, and I had actually created a task for that exactly: https://github.com/status-im/nim-status/issues/252. IMO, we should add tests for each `status/api/*.nim` module to cover the code that hasn't been unit tested in lower layers. This could be treated as an end-to-end test or a unit test, both achieving enough coverage. The lower layers should already have some unit testing, although they could use some work, but we unit test the lower layer setup and exception conversion to `Resutl[T, string]` in the higher layers. In general, and as we've discussed, we should move our tests over to using [unittest2](https://github.com/status-im/nim-unittest2), they should be normalised for consistency (ie `check` vs `assert`), and support the `setup` and `teardown` macros.

### ?? Can anyone recommend a nim test coverage lib ??

examples/client/client/events.nim
emizzle: Agreed. In the future, once we change the root `status.nim` to re-export `/status/api.nim` (which re-exports `/status/api/*.nim`), consuming libraries would only want to `import status` (see https://github.com/status-im/nim-status/pull/254#issuecomment-891583449). However, if a highly-motivated consumer wanted to utilise procs not exposed from importing `status` in itself, nothing is stopping them from importing and utilising individual modules inside of `nim-status`.

status/api/accounts.nim
emizzle: I understand what you're asking and definitely agree it needs more consideration in general. In this particular instance, `storeDerivedAccounts` is a specific abstraction that hides implementation details from the consumer. As mentioned in https://github.com/status-im/nim-status/pull/254#discussion_r681484712, highly motivated consumers could import all not re-exported modules explicitly and roll their own implementation of `storeDerivedAccounts` if so desired.

PR #259, opened 2021-08-03
feat: add /call command to TUI to execute eth rpc methods

richard-ramos:
Adds a new command `/call` or `/rpc` that receives a json string containing the call being sent to an ethereum provider
image

### DIFF
--- a/examples/client/client.nim
+++ b/examples/client/client.nim
@@ -118,3 +118,6 @@ proc addCustomToken*(self: Client, address: Address, name, symbol, color: string
 
 proc deleteCustomToken*(self: Client, index: int) {.async.} =
   asyncSpawn deleteCustomToken(self.taskRunner, status, index)
+
+proc callRpc*(self: Client, rpcMethod: string, params: JsonNode) {.async.} =
+  asyncSpawn callRpc(self.taskRunner, status, rpcMethod, params)
--- a/examples/client/client/tasks.nim
+++ b/examples/client/client/tasks.nim
@@ -7,7 +7,7 @@ import # vendor libs
   stew/byteutils
 
 import # status lib
-  status/api/[auth, tokens, wallet],
+  status/api/[auth, provider, tokens, wallet],
   status/private/[alias, protocol]
 
 import # client modules
@@ -1066,5 +1066,50 @@ proc deleteCustomToken*(index: int) {.task(kind=no_rts, stoppable=false).} =
         "error: " & e.msg, timestamp: timestamp)
       eventEnc = event.encode
       task = taskArg.taskName
+    trace "task sent event with error to host", event=eventEnc, task
+    asyncSpawn chanSendToHost.send(eventEnc.safe)
+
+proc callRpc*(rpcMethod: string, params: JsonNode) {.task(kind=no_rts, stoppable=false).} =
+  let timestamp = getTime().toUnix
+
+  if statusState != StatusState.loggedin:
+    let
+      eventNotLoggedIn = CallRpcEvent(error: "Not logged in, " &
+        "cannot call rpc methods.",
+        timestamp: timestamp)
+      eventNotLoggedInEnc = eventNotLoggedIn.encode
+      task = taskArg.taskName
+
+    trace "task sent event to host", event=eventNotLoggedInEnc, task
+    asyncSpawn chanSendToHost.send(eventNotLoggedInEnc.safe)
+    return
+
+  try:
+    let callResult = await status.callRpc(rpcMethod, params)
+    if callResult.isErr:
+      let
+        event = CallRpcEvent(error: callResult.error, timestamp: timestamp)
+        eventEnc = event.encode
+        task = taskArg.taskName
+
+      trace "task sent errored event to host", event=eventEnc, task
+      asyncSpawn chanSendToHost.send(eventEnc.safe)
+      return
+    else:
+      let
+        response = callResult.get
+        event = CallRpcEvent(response: $response, timestamp: timestamp)
+        eventEnc = event.encode
+        task = taskArg.taskName
+
+      trace "task sent event to host", event=eventEnc, task
+      asyncSpawn chanSendToHost.send(eventEnc.safe)
+  except CatchableError as e:
+    let
+      event = CallRpcEvent(error: "Error calling rpc method, " &
+        "error: " & e.msg, timestamp: timestamp)
+      eventEnc = event.encode
+      task = taskArg.taskName
+
     trace "task sent event with error to host", event=eventEnc, task
     asyncSpawn chanSendToHost.send(eventEnc.safe)
--- a/examples/client/tui/actions.nim
+++ b/examples/client/tui/actions.nim
@@ -411,3 +411,22 @@ proc action*(self: Tui, event: UserMessageEvent) {.async, gcsafe, nimcall.} =
 
     debug "TUI received user message", message, timestamp, username
     self.printMessage(message, timestamp, username, topic)
+
+# CallRpcEvent -----------------------------------------------------------------
+
+proc action*(self: Tui, event: CallRpcEvent) {.async, gcsafe,
+  nimcall.} =
+
+  # if TUI is not ready for output then ignore it
+  if self.outputReady:
+    if event.error != "":
+      self.wprintFormatError(event.timestamp, event.error)
+      return
+
+    trace "Displaying call result"
+
+    let
+      response = event.response
+      timestamp = event.timestamp
+
+    self.printResult(fmt"RPC method response: {response}", timestamp)
--- a/examples/client/tui/commands.nim
+++ b/examples/client/tui/commands.nim
@@ -1,5 +1,5 @@
 import # std libs
-  std/[sequtils, strformat, strutils, sugar]
+  std/[json, sequtils, strformat, strutils, sugar]
 
 import # client modules
   ./common, ./macros, ./screen, ./tasks
@@ -721,6 +721,43 @@ proc command*(self: Tui, command: SendMessage) {.async, gcsafe, nimcall.} =
       "client is not online, cannot send message.")
   else:
     asyncSpawn self.client.sendMessage(command.message)
+
+# Call ----------------------------------------------------------------
+
+proc help*(T: type CallRpc): HelpText =
+  let command = "call"
+  HelpText(command: command, aliases: aliased[command], parameters: @[
+    CommandParameter(name: "method", description: "RPC method"),
+    CommandParameter(name: "params", description: "Array of parameters")
+    ], description: "Calls an Ethereum RPC method")
+
+proc new*(T: type CallRpc, args: varargs[string]): T =
+  var rpcMethod = if args.len > 0: args[0] else: ""
+  var params = newJArray()
+  if args.len > 1:
+    try:
+      params = args[1..^1].join(" ").parseJson()
+    except:
+      params = newJNull()
+
+  T(rpcMethod: rpcMethod, params: params)
+
+proc split*(T: type CallRpc, argsRaw: string): seq[string] =
+  argsRaw.split(" ")
+
+proc command*(self: Tui, command: CallRpc) {.async, gcsafe,
+  nimcall.} =
+  try:
+    if command.rpcMethod == "":
+      self.wprintFormatError(getTime().toUnix,
+        "method cannot be blank, please provide an input.")
+    elif command.params.kind == JNull:
+      self.wprintFormatError(getTime().toUnix,
+        "params cannot be blank and must be a valid JSON, please provide an input.")
+    else:
+      asyncSpawn self.client.callRpc(command.rpcMethod, command.params)
+  except:
+    self.wprintFormatError(getTime().toUnix, "invalid arguments.")
 
 # Help -------------------------------------------------------------------------
 # Note: although "Help" is not alphabetically last, we need to keep this below
--- a/examples/client/tui/common.nim
+++ b/examples/client/tui/common.nim
@@ -124,6 +124,10 @@ type
   SendMessage* = ref object of Command
     message*: string
 
+  CallRpc* = ref object of Command
+    rpcMethod*: string
+    params*: JsonNode
+
 const
   TuiEvents* = [
     "InputKey",
@@ -140,6 +144,7 @@ const
     "addwalletpk": "AddWalletPrivateKey",
     "addwalletseed": "AddWalletSeed",
     "addwalletwatch": "AddWalletWatchOnly",
+    "call": "CallRpc",
     "connect": "Connect",
     "createaccount": "CreateAccount",
     "deletecustomtoken": "DeleteCustomToken",
@@ -175,6 +180,7 @@ const
     "list": "listaccounts",
     "listwallets": "listwalletaccounts",
     "part": "leavetopic",
+    "rpc": "call",
     "send": DEFAULT_COMMAND,
     "sub": "jointopic",
     "subscribe": "jointopic",
@@ -192,6 +198,7 @@ const
     "addwalletpk": @["addpk"],
     "addwalletseed": @["addseed"],
     "addwalletwatch": @["addwatch"],
+    "call": @["rpc"],
     "createaccount": @["create"],
     "deletecustomtoken": @["deletetoken"],
     "deletewalletaccount": @["delete"],
--- a/status/api.nim
+++ b/status/api.nim
@@ -1,4 +1,4 @@
 import # status modules
-  ./api/[accounts, auth, common, settings, tokens, wallet]
+  ./api/[accounts, auth, common, provider, settings, tokens, wallet]
 
-export accounts, auth, common, settings, tokens, wallet
+export accounts, auth, common, provider, settings, tokens, wallet
--- a/status/api/common.nim
+++ b/status/api/common.nim
@@ -1,15 +1,16 @@
 {.push raises: [Defect].}
 
 import # std libs
-  std/[os, typetraits]
+  std/[os, tables, typetraits]
 
 import # vendor libs
-  sqlcipher, stew/results, web3/ethtypes
+  sqlcipher, stew/results, web3, web3/ethtypes
 
 from web3/conversions as web3_conversions import `$`
 
 import # status modules
-  ../private/common, ../private/[accounts/generator/generator, database]
+  ../private/common,
+  ../private/[accounts/generator/generator, callrpc, database, settings]
 
 from ../private/conversions import parseAddress, readValue, writeValue
 from ../private/extkeys/types import Mnemonic
@@ -28,6 +29,7 @@ type
     accountsDbConn: DbConn
     dataDir*: string
     userDbConn: DbConn
+    web3Conn: Table[string, Web3]
 
 proc new*(T: type StatusObject, dataDir: string,
   accountsDbFileName: string = "accounts.sql"): T =
@@ -44,7 +46,8 @@ proc new*(T: type StatusObject, dataDir: string,
 
   let generator = Generator.new()
 
-  T(accountsDbConn: accountsDb, dataDir: dataDir, accountsGenerator: generator)
+  T(accountsDbConn: accountsDb, dataDir: dataDir, accountsGenerator: generator,
+    web3Conn: initTable[string, Web3]())
 
 proc accountsDb*(self: StatusObject): DbConn =
   self.accountsDbConn
@@ -85,3 +88,15 @@ proc initUserDb*(self: StatusObject, keyUid, password: string) {.raises:
   except DbError as e:
     # convert to a defect as we are in an unrecoverable state
     raise (ref StatusApiDefect)(parent: e, msg: "Error initializing user database")
+
+proc web3*(self: StatusObject, network: string): Web3 {.raises: [Defect, Exception]} =
+  let settings = self.userDb.getSettings()
+  if not self.web3Conn.hasKey(network):
+    self.web3Conn[network] = newWeb3(settings, network)
+  return self.web3Conn[network]
+
+proc web3*(self: StatusObject): Web3 {.raises: [Defect, Exception]}  =
+  let settings = self.userDb.getSettings()
+  if not self.web3Conn.hasKey(settings.currentNetwork):
+    self.web3Conn[settings.currentNetwork] = newWeb3(settings, settings.currentNetwork)
+  return self.web3Conn[settings.currentNetwork]
--- a/status/api/provider.nim
+++ b/status/api/provider.nim
@@ -1,0 +1,25 @@
+{.push raises: [Defect].}
+
+import # std libs
+  std/json
+
+import # vendor libs
+  chronos
+
+import # status modules
+  ../private/callrpc,
+  ./common
+
+type
+  CallRpcResult* = Result[JsonNode, string]
+
+proc callRpc*(self: StatusObject, rpcMethod: string, params: JsonNode):
+  Future[CallRpcResult] {.async.} =
+
+  if not self.isLoggedIn:
+    return CallRpcResult.err "Not logged in. Must be logged in"
+
+  try:
+    return CallRpcResult.ok await self.web3.callRpc(rpcMethod, params)
+  except CatchableError as e:
+    return CallRpcResult.err e.msg
--- a/status/private/callrpc.nim
+++ b/status/private/callrpc.nim
@@ -58,36 +58,44 @@ type
     net_listening = "net_listening"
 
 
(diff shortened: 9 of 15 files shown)


examples/client/tui/commands.nim
```diff
@@ -722,6 +722,72 @@ proc command*(self: Tui, command: SendMessage) {.async, gcsafe, nimcall.} =
...
+# Call ----------------------------------------------------------------
+
+proc help*(T: type CallRPC): HelpText =
+  let command = "call"
+  HelpText(command: command, aliases: aliased[command], parameters: @[
+    CommandParameter(name: "input", description: "Options JSON")
+    ], description: "Calls an Ethereum RPC method")
+
+proc new*(T: type CallRPC, args: varargs[string]): T =
+  var
```
siphiuel, line 734: Not sure if this is better, but it's possible to write `let input = if args.len > 0: args[0] else: ""`

siphiuel (approved):

examples/client/client/tasks.nim
```diff
@@ -1047,3 +1047,53 @@ proc deleteCustomToken*(index: int) {.task(kind=no_rts, stoppable=false).} =
       task = taskArg.taskName
     trace "task sent event with error to host", event=eventEnc, task
     asyncSpawn chanSendToHost.send(eventEnc.safe)
+
+proc callRPC*(input: string) {.task(kind=no_rts, stoppable=false).} =
```
emizzle, line 1051:
```suggestion
proc callRPC*(input: JsonNode) {.task(kind=no_rts, stoppable=false).} =
```

examples/client/tui/common.nim
```diff
@@ -124,6 +124,14 @@ type
   SendMessage* = ref object of Command
     message*: string
 
+  CallRPC* = ref object of Command
+    input*: string
```
emizzle, line 128:
```suggestion
  CallRpc* = ref object of Command
    input*: JsonNode
```

status/api/provider.nim
```diff
@@ -0,0 +1,34 @@
+import # std libs
+  json
+
+import # nim-status libs
+  ./common, ../settings, ../callrpc, chronicles
+
+type
+  CallRPCResult* = Result[JsonNode, string]
+
+proc callRPC*(self: StatusObject, input: string): CallRPCResult  {.raises:
```
emizzle, line 10:
```suggestion
proc callRPC*(self: StatusObject, input: JsonNode): CallRPCResult  {.raises:
```

examples/client/tui/commands.nim
```diff
@@ -722,6 +722,72 @@ proc command*(self: Tui, command: SendMessage) {.async, gcsafe, nimcall.} =
...
+    input = ""
+  if args.len > 0:
+    input = args[0]
+  T(input: input)
```
emizzle, line 738: Maybe we could have `input` be `JsonNode`, such that we are forced to validate correct JSON input here. Especially if the user is typing in JSON 😳  If the JSON can't be parsed or is invalid, it would prevent a round-trip through the task runner.

emizzle (changes requested):
Overall, really nice work @richard-ramos 🤩

### General comment
As a general comment, given that this PR implements only part of send transaction, I think that we could treat this PR in one of two ways:
1. as a draft until `sendTransaction` is included
2. as a PR but without `sendTransaction` and follow up with a `sendTransaction` PR

### Using ethcallsigs
I really like the use the use of `nim-web3`, well done.

For calling RPC methods, constructing JSON calls and then pasting them in to the TUI feels like it is a bit on the unfriendly side, but I do see how this functionality could be useful for debugging in the future. I guess in an ideal world, it could potentially be more user-friendly to parse parameters based on the method. As an example, if we wanted to get a block by block number, `eth_getBlockByNumber`, we could do something like:
```shell
# tui input
/call eth_getBlockByNumber 0x123 true
```
These parameters could then be parsed and plugged in to the appropriate [ethcallsig](https://github.com/status-im/nim-web3/blob/master/web3/ethcallsigs.nim), with their usage being like so:
```
# status/api/provider.nim
let block = await self.web3.provider.eth_getBlockByNumber(blockId, fullTransactions)
```
The parsing for each call would have to be different. Compared to `eth_getBlockByNumber`, `eth_getStorageAt` needs three different parameters:
```shell
# tui input
/call eth_getStorageAt 0xBdc1F0460b38876a82Fe8416210DbC49e28c6298 10 0x1234
```
```
# status/api/provider.nim
let block = await self.web3.provider.eth_getStorageAt(address, quantity, blockId)
```
It's worth noting that in other parts of `nim-status` (future additions, out of scope of this PR), if we need to call RPC methods, we should be using the [ethcallsigs](https://github.com/status-im/nim-web3/blob/master/web3/ethcallsigs.nim) instead of unnecessarily constructing JSON objects that skirt around compiler support. Likely, these use cases would not need to be abstracted out in to a generalised call like we are doing for `callRpc`, so hopefully should be much easier to implement ad-hoc.

### Cosmetic preferences
To summarise a general comment about casing, I prefer to treat acronyms like any other word and therefore use camel casing rules, ie I prefer `callRpcEvent` as opposed to `callRPCEvent`. Any usage of `RPC` would be `rpc`, unless it is not at the start, in which case it would be `Rpc`. Of course, this is only my opinion though. IMO, it seems for long acronyms like `secp256k1`, it could feel like someone was swearing at you `SECP256K1` 🤬 😂

iurimatias:
> Overall, really nice work @richard-ramos 🤩
> 
> ### General comment
> As a general comment, given that this PR implements only part of send transaction, I think that we could treat this PR in one of two ways:
> 
> 1. as a draft until `sendTransaction` is included
> 2. as a PR but without `sendTransaction` and follow up with a `sendTransaction` PR
> 
> ### Using ethcallsigs
> I really like the use the use of `nim-web3`, well done.
> 
> For calling RPC methods, constructing JSON calls and then pasting them in to the TUI feels like it is a bit on the unfriendly side, but I do see how this functionality could be useful for debugging in the future. I guess in an ideal world, it could potentially be more user-friendly to parse parameters based on the method. As an example, if we wanted to get a block by block number, `eth_getBlockByNumber`, we could do something like:
> 
> ```shell
> # tui input
> /call eth_getBlockByNumber 0x123 true
> ```
> 
> These parameters could then be parsed, validated, and plugged in to the appropriate [ethcallsig](https://github.com/status-im/nim-web3/blob/master/web3/ethcallsigs.nim), with their usage being like so:
> 
> ```
> # status/api/provider.nim
> let block = await self.web3.provider.eth_getBlockByNumber(blockId, fullTransactions)
> ```
> 
> The parsing and validating for each call would have to be different due to the different parameters. Compared to `eth_getBlockByNumber`, `eth_getStorageAt` needs three different parameters:
> 
> ```shell
> # tui input
> /call eth_getStorageAt 0xBdc1F0460b38876a82Fe8416210DbC49e28c6298 10 0x1234
> ```
> 
> ```
> # status/api/provider.nim
> let block = await self.web3.provider.eth_getStorageAt(address, quantity, blockId)
> ```
> 
> But it feels like when we are allowing free text, we could be validating parameters for better UX and then ultimately that would provide for better compiler support with strongly-typed access.
> 
> It's worth noting that in other parts of `nim-status` (future additions, out of scope of this PR), if we need to call RPC methods, we should be using the [ethcallsigs](https://github.com/status-im/nim-web3/blob/master/web3/ethcallsigs.nim) instead of unnecessarily constructing JSON objects that skirt around compiler support. Likely, these use cases would not need to be abstracted out in to a generalised call like we are doing for `callRpc`, so hopefully should be much easier to implement ad-hoc.
> 
> ### Cosmetic preferences
> To summarise a general comment about casing, I find it preferable to treat acronyms like any other word and therefore use camel casing rules, ie using `callRpcEvent` as opposed to `callRPCEvent`. Any usage of `RPC` would be `rpc`, unless it is not at the start, in which case it would be `Rpc`. Of course, this is only my opinion though.

- we should avoid drafts if we can, specially in a project like this, it risks having a hard rebase needed, better to do in a separate PR
- the unfriendly TUI thing is not something we need to be worried about and it's not worth the work at this stage, we're building a library and the TUI is there just to help but it's not meant to be a user friendly client (not at this stage anyway)

status/callrpc.nim
```diff
@@ -52,21 +52,28 @@ type RemoteMethod* {.pure.} = enum
   net_listening = "net_listening"
 
 
-proc newWeb3*(settings: Settings): Web3 {.raises: [CatchableError, Defect,
-  ref Web3Error].} =
-
-  let network = settings.getCurrentNetwork()
+proc getWeb3Conn(networkName: string, network: Option[Network]): Web3 {.raises:
```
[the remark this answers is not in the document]
richard-ramos: This is done this way so I can pass the result of `settings.getNetwork(networkName)` and `settings.getCurrentNetwork()` directly to the `getWeb3Conn` and have this proc check if the network contains a value,  opossed to having this verification done in both `newWeb3` procs. 
Also `getWeb3Conn` is not a procedure that can be exported. it's a helper for `newWeb3`. Perhaps it should have a different name to avoid this confusion?

status/networks.nim
```diff
@@ -24,7 +24,7 @@ type
     name*: string
 
 # set via `nim c` param `-d:INFURA_TOKEN:[token]`; should be set in CI/release builds
-const INFURA_TOKEN {.strdefine.} = ""
+const INFURA_TOKEN {.strdefine.} = "220a1abb4b6943a093c35d0ce4fb0732" # @TODO: remove this
```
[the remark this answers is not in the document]
richard-ramos: It should be okay. I copied it from https://github.com/status-im/status-desktop/blob/master/Makefile#L220

status/networks.nim
```diff
@@ -41,7 +41,7 @@ const DEFAULT_NETWORKS* = @[
       dataDir: "/ethereum/testnet_rpc",
       upstreamConfig: UpstreamConfig(
         enabled: true,
-        url: "https://ropsten.infura.io/v3/" & INFURA_TOKEN
+        url: "wss://ropsten.infura.io/ws/v3/" & INFURA_TOKEN
```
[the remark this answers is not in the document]
richard-ramos: `https://` was not working. I had to switch to websockets.

examples/client/tui/commands.nim
```diff
@@ -722,6 +722,72 @@ proc command*(self: Tui, command: SendMessage) {.async, gcsafe, nimcall.} =
...
+proc new*(T: type SendTransaction, args: varargs[string]): T =
+  var
+    input = ""
+    password = ""
+  if args.len > 0:
+    input = args[0]
+  if args.len > 1:
+    password = args[1]
+
+  T(input: input, password: password)
```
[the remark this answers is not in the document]
richard-ramos: Yeah. Sounds like a good idea

richard-ramos:
I removed all the code related to sending transactions because it will be part of a separate PR.

I also did most of the required changes but in relation to what was asked in "Using ethcallsigs", I decided to implement it differently: The call command now receives two parameters, with the last one being optional, line in these examples
```
/call eth_gasPrice
/call eth_getBlockByNumber ["0x5BAD55",false]
```

While I agree that using ethcallsigs make sense to use in other parts of the project, I do think that for the TUI, implementing it this way should be good enough for now to achieve a middle point between introducing a JSON which is not great for UX, and having validation per parameters depending on the RPC method being called which would probably require to allocate a lot of time for something that IMO will not be used IRL

examples/client/tui/commands.nim
```diff
@@ -722,6 +722,43 @@ proc command*(self: Tui, command: SendMessage) {.async, gcsafe, nimcall.} =
...
+
+proc command*(self: Tui, command: CallRpc) {.async, gcsafe,
+  nimcall.} =
+  try:
+    if command.rpcMethod == "":
+      self.wprintFormatError(getTime().toUnix,
+        "method cannot be blank, please provide an input.")
+    elif command.params.kind == JNull:
+      self.wprintFormatError(getTime().toUnix,
+        "params cannot be blank, please provide an input.")
```
emizzle, line 756:
```suggestion
        "params cannot be blank and must be valid JSON, please provide an input.")
```

emizzle: This is because if there was a JSON parse error in `new` (ie user entered bad json), then this will be shown.

status/callrpc.nim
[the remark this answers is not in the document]
emizzle: Ah I see, I had forgotten that `getNetwork` and `getCurrentNetwork` returned `Option[Network]` 👍

status/networks.nim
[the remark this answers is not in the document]
emizzle: Oh that's very interesting. Is infura removing support for https? (no need to answer if you don't know, just asking out loud).

emizzle (approved):
Awesome work, Richard! Thanks for all the updates 💪

emizzle:
@iurimatias 

> we should avoid drafts if we can, specially in a project like this, it risks having a hard rebase needed, better to do in a separate PR

In my opinion, draft PRs are a great way for a developer to welcome feedback on a PR that isn't quite polished and ready for merging. This can help ensure they are going in the right direction before the time is spent cleaning things up and getting it ready to merge. What is a hard rebase?

> the unfriendly TUI thing is not something we need to be worried about and it's not worth the work at this stage, we're building a library and the TUI is there just to help but it's not meant to be a user friendly client (not at this stage anyway)

I understand and definitely agree with not wanting to put too much effort in to the TUI. I wasn't really expecting the work on the TUI to get done; a lot of it was commentary on idealities, but also a note on how we could move forward when calling RPC methods *from within nim-status* in the future.  I think the `JsonNode` input parameters is a good solution for the `callRpc` API proc. And richard's changes to the tui were a perfect middle ground effort.

examples/client/tui/actions.nim
```diff
@@ -411,3 +411,22 @@ proc action*(self: Tui, event: UserMessageEvent) {.async, gcsafe, nimcall.} =
 
     debug "TUI received user message", message, timestamp, username
     self.printMessage(message, timestamp, username, topic)
+
+# CallEvent --------------------------------------------------------------------
```
michaelsbradleyjr, line 415: Not a big deal, but it's nice if the name in the comment matches the type name.

Actually, at this point it probably makes to create an `acitons/` dir and start splitting actions into separate modules. That can be done in a separate PR, of course; I'm just thinking aloud.

examples/client/tui/commands.nim
```diff
@@ -722,6 +722,43 @@ proc command*(self: Tui, command: SendMessage) {.async, gcsafe, nimcall.} =
...
+    CommandParameter(name: "method", description: "RPC method"),
+    CommandParameter(name: "params", description: "Array of parameters")
+    ], description: "Calls an Ethereum RPC method")
+
+proc new*(T: type CallRpc, args: varargs[string]): T =
+  var rpcMethod = if args.len > 0: args[0] else: ""
+  var params = newJArray()
+  if args.len > 1:
+    try:
+      params = args[1].parseJson()
```
michaelsbradleyjr, line 739: The user might have typed in valid json that contained spaces, e.g. between `"key":` and the following value.

So here I think `args[1..^1]` should be joined on single-space (`" "`) and then parsed.

Also, I think we should have an `else` case for `if args.len > 1` where params is set to the empty string.

Then, in `proc command` below for `command: CallRpc`, the situation re: blank params and invalid JSON can be separated.

examples/client/tui/commands.nim
```diff
@@ -722,6 +722,43 @@ proc command*(self: Tui, command: SendMessage) {.async, gcsafe, nimcall.} =
...
+
+proc command*(self: Tui, command: CallRpc) {.async, gcsafe,
+  nimcall.} =
+  try:
+    if command.rpcMethod == "":
+      self.wprintFormatError(getTime().toUnix,
+        "method cannot be blank, please provide an input.")
+    elif command.params.kind == JNull:
+      self.wprintFormatError(getTime().toUnix,
+        "params cannot be blank and must be a valid JSON, please provide an input.")
```
michaelsbradleyjr, line 756: See my previous comment re: separating `params cannot be blank` and `valid JSON`.

examples/client/tui/common.nim
```diff
@@ -192,6 +198,7 @@ const
     "addwalletpk": @["addpk"],
     "addwalletseed": @["addseed"],
     "addwalletwatch": @["addwatch"],
+    "call": @["call"],
```
michaelsbradleyjr, line 201:
```suggestion
    "call": @["rpc"],
```

status/api.nim
```diff
@@ -1,4 +1,4 @@
 import # status modules
-  ./api/[accounts, auth, common, settings, tokens, wallet]
+  ./api/[accounts, auth, common, settings, tokens, wallet, provider]
```
michaelsbradleyjr, line 2: Again, it's nice to consistently alphabetize.

status/api.nim
```diff
@@ -1,4 +1,4 @@
...
 
-export accounts, auth, common, settings, tokens, wallet
+export accounts, auth, common, settings, tokens, wallet, provider
```
michaelsbradleyjr, line 4: Here also.

status/api/common.nim
```diff
@@ -1,13 +1,13 @@
...
-  std/[os, typetraits]
+  std/[os, tables, typetraits]
 
 import # vendor libs
-  sqlcipher, stew/results
+  sqlcipher, stew/results, web3
 
 import # status modules
-  ../private/[accounts/generator/generator, database]
+  ../private/[accounts/generator/generator, database, callrpc, settings]
```
michaelsbradleyjr, line 10: Here also.

status/api/provider.nim
```diff
@@ -0,0 +1,19 @@
+import # std libs
+  json
+
+import # status modules
+  ../private/[callrpc],
```
michaelsbradleyjr, line 5: If there's only one, the `[]` can be left out, but no big deal.

status/api/common.nim
```diff
@@ -57,3 +59,15 @@ proc initUserDb*(self: StatusObject, keyUid, password: string) {.raises:
   [Exception].} =
 
   self.userDbConn = initializeDB(self.dataDir / keyUid & ".db", password)
+
+proc web3*(self: StatusObject, network: string): Web3 {.raises: [Exception]} =
```
michaelsbradleyjr, line 63:
```suggestion
proc web3*(self: StatusObject, network: string): Web3 {.raises: [Defect, Exception]} =
```

We should indicate `Defect` in the `{.raises.}` pragma.

It would better to not declare `Exception` in the `{.raises.}` pragma, as it somewhat defeats the purpose (even if it makes the compiler happy). What we really want is to declare specific kinds of exceptions, e.g. `KeyError` or maybe error/s that getSettings could raise.

However, there's a big refactor @emizzle is working on (almost done) related to this (in a general sense), so it's something that can be sorted out in a future PR if [`KeyError`](https://nim-lang.org/docs/system.html#KeyError) isn't enough.

status/private/callrpc.nim
```diff
@@ -57,28 +57,35 @@ type RemoteMethod* {.pure.} = enum
...
 
-proc newWeb3*(settings: Settings): Web3 {.raises: [CatchableError, Defect,
-  ref Web3Error].} =
-
-  let network = settings.getCurrentNetwork()
+proc getWeb3Conn(networkName: string, network: Option[Network]): Web3 {.raises: 
+  [CatchableError, Defect, ref Web3Error].} =
   if network.isNone:
-    raise (ref Web3Error)(msg: "config not found for network " & settings.currentNetwork)
+    raise (ref Web3Error)(msg: "config not found for network" & networkName)
```
michaelsbradleyjr, line 63:
```suggestion
    raise (ref Web3Error)(msg: "config not found for network " & networkName)
```

status/private/networks.nim
```diff
@@ -25,7 +25,7 @@ type
     name*: string
 
 # set via `nim c` param `-d:INFURA_TOKEN:[token]`; should be set in CI/release builds
-const INFURA_TOKEN {.strdefine.} = ""
+const INFURA_TOKEN {.strdefine.} = "220a1abb4b6943a093c35d0ce4fb0732" # @TODO: remove this
```
michaelsbradleyjr, line 28: We can eventually supply the default via the Makefile, like we do in status-desktop.

michaelsbradleyjr (changes requested):
Really nice work!

I've offered a few stylistic points below, but they're not blockers for this PR.

The two blockers concern (1) my comments re: parsing of the JSON supplied by the user via the TUI, and (2) the mistake in the `aliased` table.

michaelsbradleyjr:
> the unfriendly TUI thing is not something we need to be worried about and it's not worth the work at this stage, we're building a library and the TUI is there just to help but it's not meant to be a user friendly client (not at this stage anyway)

I agree, the focus should be on dogfooding and *not* great UI/UX.

But, echoing @emizzle's sentiments, I think we can strike a balance between landing too-poor UI/UX and spending unwarranted time on UI/UX. To my mind, @richard-ramos performed an *awesome* balancing act in this PR.

> we should avoid drafts if we can, specially in a project like this, it risks having a hard rebase needed, better to do in a separate PR

A few thoughts, and some *only-my-opinions* are definitely included:

This week it would have been quite difficult for one PR or another to avoid a rebase. We've not been moving stuff around for the sake of it, but trying to make the foundations more solid, and some of those efforts have been happening in parallel, so any one PR would likely have to take the changes into account.

The reorgs/refactors won't go on forever, we're very close to having the foundation we need and want.

Re: this repository and its current GHA workflows, putting a PR in *Draft* is a way for the author to signal:
* I'm want this branch to be built/tested on all three platforms (Linux, macOS, Windows).
* I'm inviting review of the work I've done, which I feel needs some feedback before it's ready to merge.

We quite recently ran into a situation where Eric and I decided to merge one of @siphiuel's PRs that wasn't ready yet, though that wasn't clear to us. While we could use an approach of author/other marking a PR with a red-colored `don't merge yet` label, putting a PR into draft seems like a more elegant solution, since draft PRs simply cannot be merged.

Some of us (sometimes) build on all three platforms locally, in the context of PR authoring, but most of us don't or don't do it all the time. We *could* change nim-status' GHA workflows so they run on any commit to any branch, but we don't have an unlimited number of GHA minutes per month (for the status-im org), so limiting GHA runs to PRs and the `master` branch seems like a good idea.

examples/client/client.nim
```diff
@@ -118,3 +118,6 @@ proc addCustomToken*(self: Client, address: Address, name, symbol, color: string
 
 proc deleteCustomToken*(self: Client, index: int) {.async.} =
   asyncSpawn deleteCustomToken(self.taskRunner, status, index)
+
+proc callRpc*(self: Client, rpcMethod: string, params: JsonNode) {.async.} =
```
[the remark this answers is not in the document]
emizzle: There is actually already an enum in `status/callrpc.nim`: `RemoteMethod`. Maybe we could use that?

examples/client/tui/commands.nim
michaelsbradleyjr: > `else` case for if `args.len > 1` where params is set to the empty string.

Whoops, looking over it again I see params needs to be of type `JsonNode`. So there's no need for an `else` case, but we could still separate the cases w.r.t. checking if it's JNull else checking the `len` of JArray being `> 0`.

michaelsbradleyjr (approved):

michaelsbradleyjr:
@richard-ramos I just rebased this PR against current `master`. In the process of doing so, I realized that nim-status' `callRpc` is synchronous but it should be async.

So I introduced and squashed-in changes that make `callRpc` async. I was careful to review the rebased PR + my changes relative to the changes the original PR made on top of 1adb3045d66add8df44dad7a21cd1fb8e0be4dcf. Tests pass, the client builds and `/callRpc` works as expected. Even so, you may want to take a look to check that I didn't mess something up in the process.

In `api/provider` there's a `raises: [Exception]`. Now that #265 has landed, we should probably take a different approach for that. @emizzle, could you take a look?

michaelsbradleyjr:
I've temporarily added a `don't merge yet` label until Richard has had time to check the changes I introduced and we together decided what we want to do re: the `raises: [Exception]`.

emizzle:
@michaelsbradleyjr @richard-ramos, I've updated (and squashed-in) this brach to fix the `{.raises: [Exception].}.` The `api/proivder.nim` module didn't have a `{.push raises: [Defect].}` at the top of the module, so the `{.raises: [Exception].}.` likely was satisfying the compiler's analysis that a `Defect` could have been raised. 

Fixed by adding `{.push raises: [Defect].}` and removing the `{.raises.}` pragma from the proc (effectively equivalent to `{.raises: [Defect].}`).

PR #260, opened 2021-08-04
refactor: organize non-api modules under status/private/

michaelsbradleyjr:
Also clean up import sections in all modules under examples/, status/, and test/


emizzle (approved):
Very nice, thank you for doing this!

siphiuel (approved):

PR #261, opened 2021-08-04
refactor: example client does not import non-api modules of nim-status

michaelsbradleyjr:
There's one remaining import of a non-api (private) module in `examples/client/client/tasks.nim`:

```
../../status/private/protocol
```

That one will be removed in a forthcoming PR that moves import/usage of nim-waku from the example client to the nim-status library itself.

### DIFF
--- a/examples/client/client/common.nim
+++ b/examples/client/client/common.nim
@@ -5,12 +5,12 @@ import # vendor libs
   task_runner
 
 import # status lib
-  ../../../status/private/accounts/public_accounts
+  ../../../status/api/accounts
 
 import # client modules
   ../config
 
-export config, public_accounts, sets, strformat, strutils, task_runner, times
+export accounts, config, sets, strformat, strutils, task_runner, times
 
 logScope:
   topics = "client"
--- a/examples/client/client/events.nim
+++ b/examples/client/client/events.nim
@@ -2,7 +2,7 @@ import # vendor libs
   web3/ethtypes
 
 import # status lib
-  ../../../status/[api, private/tokens]
+  ../../../status/api/[tokens, wallet]
 
 import # client modules
   ./common
--- a/examples/client/client/tasks.nim
+++ b/examples/client/client/tasks.nim
@@ -7,14 +7,13 @@ import # vendor libs
   stew/byteutils
 
 import # status lib
-  ../../status/api,
-  ../../status/private/accounts/[accounts, public_accounts],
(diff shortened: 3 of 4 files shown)


examples/client/client/common.nim
```diff
@@ -5,12 +5,12 @@ import # vendor libs
   task_runner
 
 import # status lib
-  ../../../status/private/accounts/public_accounts
+  ../../../status/api/accounts
```
emizzle, line 8: Shouldn't this now be available via `import status/api/accounts`?

emizzle (commented):
I'm likely missing something, but shouldn't we be able to import directly from `status/api/<module>`? As opposed to accessing via relative path?

examples/client/client/common.nim
michaelsbradleyjr: We have to use a relative path for `status` in modules under `examples/` for the same reason we have to do so for modules under `test/`.

The nimbus-build-system (its `env.sh` plus work it does when invoking `make` targets) allows for the Nim compiler / nimble to load the modules under `vendor/` as if they were installed in a conventional nimble setup. But apart from dependencies and transitive nimbus-build-system dependencies under `vendor/`, it does nothing for us with regard to `examples/` and `test/` in relation to `status.im` / `status/` in the root of the repo.

michaelsbradleyjr: That being said, if someone using the nimbus-build-system installed nim-status in their `vendor/` (or it was one of their transitive deps under `vendor/`), then they would be able to `import status` or `import status/api/accounts`, etc.

emizzle: Makes complete senese. I was thinking as this is an example of client consuming `nim-status`, that it serves to show potentials consumers of `nim-status` (as an external dependency using the nimbus-build-system) how they could simply `import status/api/accounts`. Maybe it might be worth a comment in the source so that devs looking to implement their own client will know they can do it that way if they weren't aware. Of course this is something we can put in the README as well.

emizzle (approved):

examples/client/client/common.nim
michaelsbradleyjr: I filed an enhancement issue for nimbus-build-system: https://github.com/status-im/nimbus-build-system/issues/24.

michaelsbradleyjr: See PR #262.

🚀 Thanks @stefantalpalaru!

PR #262, opened 2021-08-06
refactor: create nim.cfg in examples/ and test/ so `status` lib can be imported unprefixed

michaelsbradleyjr:


emizzle (approved):
Nice work!! I'm still a bit in the dark on how `nim.cfg` helps to resolve the path correctly. does the `--path: ../` mean that the `nim-build-system` will search `../vendor` for `nim-status`?

Also, I think it's probably fine to keep the tests as `../status/...` (as they were before). If we include tests for the client at some point, we would want to have those `import status`.

michaelsbradleyjr:
> Nice work!! I'm still a bit in the dark on how `nim.cfg` helps to resolve the path correctly. Does the `--path: ../` mean that the `nim-build-system` will search `../vendor` for `nim-status`?

It means that nimble will search `../`,  relative to `examples/` or `test/`, for `status`.

> Also, I think it's probably fine to keep the tests as `../status/...` (as they were before). If we include tests for the client at some point, we would want to have those `import status`.

With the `test/nim.cfg` in place in this PR, in modules under `test/` we should already be able to `import examples/client/client/baz` and create tests for such a module.

So, given the convenience factor, it seems to me that:

```
import status/api/foo
```
or
```
import status/private/foo/bar
```
or
```
import examples/client/client/baz
```
within modules under `test/` is a nice convenience.

OTOH, I may have misunderstood something, so please comment further if something seems problematic or less than ideal.

PR #265, opened 2021-08-08
feat: add exception handling

emizzle:
Closes: #239

Remove all `except Exception as e` where possible. In some instances, transitive dependencies out of our control were the source of this error. Even in instances of the Nim library, like `parseJson`, for example, force us to `except Exception` to prevent it from bubbling up to higher layers in the abstraction.

At the lowest levels, we are catching all errors that inherit from `CatchableError` and translating them to be useful for layers above. At the highest level, which is our public API layer, meant to be directly consumed by consumer clients, we are catching all catchable exceptions from the layers below and translating them in to a Result type.

Developer guid documentation for exception handling that explains the motivation for these changes in its own [developer guide](https://hackmd.io/rkpbKsEYT_umNlHU1IcbiQ)

### DIFF
--- a/status/api/accounts.nim
+++ b/status/api/accounts.nim
@@ -1,45 +1,67 @@
 {.push raises: [Defect].}
 
 import # std libs
-  std/[json, tables, times]
+  std/[json, re, tables, times]
 
 import # vendor libs
   eth/keyfile/uuid, secp256k1
 
 import # status modules
-  ../private/accounts/[accounts, generator/generator, public_accounts],
+  ../private/accounts/[accounts, public_accounts],
+  ../private/accounts/generator/[account, generator],
   ../private/[alias, conversions, identicon, settings],
   ../private/extkeys/[paths, types],
   ./common
 
 export
-  accounts, alias, common, conversions, generator, identicon, paths,
-  public_accounts, secp256k1, settings, types, uuid
+  common, public_accounts
+  # TODO: are these exports needed?
+  #   accounts, alias, conversions, generator, identicon, paths,
+  #   public_accounts, secp256k1, settings, types, uuid
 
 type
   PublicAccountResult* = Result[PublicAccount, string]
 
+  PublicAccountsResult* = Result[seq[PublicAccount], string]
+
+  ChatAccountResult* = Result[accounts.Account, string]
+
 proc storeDerivedAccounts(self: StatusObject, id: UUID, keyUid: string,
   paths: seq[KeyPath], password, dir: string,
-  accountType: AccountType): PublicAccountResult {.raises: [Defect,
-  ref IOError, ref KeyError, ref OSError, SqliteError, ValueError].} =
+  accountType: AccountType): PublicAccountResult =
+
+  let accountInfos = ?self.accountsGenerator.storeDerivedAccounts(id, paths,
+      password, dir)
+  var whisperAcct: AccountInfo
+  try:
+    whisperAcct = accountInfos[PATH_WHISPER]
+  except KeyError as e:
+    return PublicAccountResult.err "Error getting whisper account from " &
+      "derived accounts: " & e.msg
 
   let
-    accountInfos = ?self.accountsGenerator.storeDerivedAccounts(id, paths,
-      password, dir)
-    whisperAcct = accountInfos[PATH_WHISPER]
+    alias = try: whisperAcct.publicKey.generateAlias
+            except RegexError as e:
+              return PublicAccountResult.err "Error generating alias for " &
+                "public key: " & e.msg
     pubAccount = PublicAccount(
       creationTimestamp: getTime().toUnix,
-      name: whisperAcct.publicKey.generateAlias(),
+      name: alias,
       identicon: whisperAcct.publicKey.identicon(),
       keycardPairing: "",
       keyUid: keyUid # whisper key-uid
     )
-
-  self.accountsDb.saveAccount(pubAccount)
+  try:
+    self.accountsDb.saveAccount(pubAccount)
+  except PublicAccountDbError as e:
+    return PublicAccountResult.err "Failed to store derived accounts: " & e.msg
 
   let
-    defaultWalletAccountDerived = accountInfos[PATH_DEFAULT_WALLET]
+    defaultWalletAccountDerived =
+      try: accountInfos[PATH_DEFAULT_WALLET]
+      except KeyError as e:
+        return PublicAccountResult.err "Error getting default wallet from " &
+          "derived accounts: " & e.msg
     defaultWalletPubKeyResult =
       SkPublicKey.fromHex(defaultWalletAccountDerived.publicKey)
     whisperAccountPubKeyResult =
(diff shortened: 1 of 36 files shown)


status/api/accounts.nim
```diff
@@ -90,13 +114,18 @@ proc storeDerivedAccounts(self: StatusObject, id: UUID, keyUid: string,
...
+    return PublicAccountResult.err errorMsg & e.msg
+  except ValueError as e:
+    return PublicAccountResult.err "Error parsing address of default wallet " &
+      "account: " & e.msg
 
 
 proc createAccount*(self: StatusObject, mnemonicPhraseLength: int,
   bip39Passphrase, password: string, dir: string): PublicAccountResult {.raises:
-  [ref Exception, ref ValueError].} =
+  [Defect, StatusApiError].} =
```
arnetheduck, line 128: another small detail worth considering: exceptions and `Result` are both error handling mechanisms that by and large address the same problem - we tend to avoid using both in the same function, else the calling code must go through some gymnastics to handle both types - one case were this potentially doesn't apply is if there's some truly exceptional construct that _should_ ignore all "normal" error handling and jump straight to an exception handler, but that's .. rare.

emizzle: Very good point and I did strive for that. Looks like this one slipped through. Will fix so that the StatusApiError is caught.

michaelsbradleyjr: Making sure I understand... so if the bases are covered inside the proc, then we don't need/want the raises pragma here (keeping in mind that at the top of the module we have `{.push raises: [Defect].}`)?

arnetheduck: > we don't need/want the raises pragma here

correct

emizzle: I believe I did catch StatusApiError in other top-level, public-facing procs. Really think I just missed this one. May be others as well, will review.

michaelsbradleyjr:
Tests were failing locally for me, just like in CI. I added a WIP commit that adds Defect to the raises pragma for initializeDB, and now tests are passing locally and in CI.

status/api/wallet.nim
```diff
@@ -25,82 +27,106 @@ type
...
+    return WalletAccountResult.err errorMsgSetting & e.msg
+  except UnpackError as e:
+    return WalletAccountResult.err errorMsgSetting & e.msg
+  except ValueError as e:
+    return WalletAccountResult.err errorMsgParse & e.msg
 
 proc storeImportedWalletAccount(self: StatusObject, privateKey: SkSecretKey,
-  name, password, dir: string, accountType: AccountType): WalletAccountResult =
+  name, password, dir: string, accountType: AccountType): WalletAccountResult
+  {.raises: [].} =
```
michaelsbradleyjr, line 129: Is it important for some procs to basically undo what `{.push raises: [Defect].}` does at the top of the module?

status/private/chats.nim
```diff
@@ -1,186 +1,205 @@
 {.push raises: [Defect].}
 
 import # std libs
-  std/[json, marshal, options, strformat]
+  json, marshal, options, sequtils, strformat
```
michaelsbradleyjr, line 4:
```suggestion
  std/[json, marshal, options, sequtils, strformat]
```

status/private/identicon.nim
```diff
@@ -45,62 +46,66 @@ proc generate(id: string): Identicon {.raises: [].} =
...
+    let strm = newStringStream()
+    png.writeChunks(strm)
+    strm.flush()
+    strm.setPosition(0)
+    let encoded = base64.encode(strm.readAll())
+    strm.close()
+    return "data:image/png;base64," & encoded
+  except Exception as e:
+    raise (ref IdenticonError)(parent: e, msg: "Error generating rendering " &
+      "base64 identicon")
```
michaelsbradleyjr, line 106:
```suggestion
    raise (ref IdenticonError)(parent: e, msg: "Error generating base64 " &
      "encoded identicon")
```

status/private/migration.nim
```diff
@@ -1,55 +1,88 @@
...
+  MigrationResult* = Result[Migration, string]
 
+  MigrationError* = object of StatusError
 
-proc createMigrationTableIfNotExists*(db: DbConn) {.raises: [Exception].} =
+
+proc createMigrationTableIfNotExists*(db: DbConn) {.raises: [MigrationError].} =
 
   var migration: Migration
+  const errorMsg ="Error creating migration table if not exists"
```
michaelsbradleyjr, line 28:
```suggestion
  const errorMsg = "Error creating migration table if not exists"
```

status/private/tx_history.nim
```diff
@@ -393,114 +519,143 @@ var fetchLock: RLock
...
+              fetchTxDetails(address, txToData)
+
+            let transferInfo = TransferInfo(
+              address: address,
+              balance: balance,
+              blockNumber: lastBlockNumber,
+              txCount: txCount)
+            writeToDb(TxDbData(info: transferInfo, txToData: txToData))
+
+      sleep(schedulerInterval)
```
michaelsbradleyjr, line 586: I realize this code didn't originate with this PR, but we can't `sleep`, as that would block the thread on which nim-status is running. It should be up to a client implementor (e.g. ourselves with the example client) to fetch history in a loop using `sleepAsync`.

Also, as seen later in this module, this logic involves a lock and and an extra thread, which is also a no-go. I'ved created an issue re: this module needing to be overhauled: #266.

michaelsbradleyjr (commented):
Awesome work, what a monumental effort!  I had to take breaks while reviewing it. 😅 

Below I have a couple of questions and some minor style-related adjustments. I also made some comments re: problems in `status/private/tx_history.nim`, but fixing those is definitely a separate task.

Note my comment above re: adding a WIP commit; should be squashed into the main commit before merge.

Also, given the huge scope of this PR, I'm sure there are things here and there that could be touched up and I didn't notice them. But the fact it compiles correctly, the tests all pass, and the example client compiles and runs correctly — that's all a testament to your fortitude and thorough work on it. Thank you for tackling this!

status/api/accounts.nim
```diff
@@ -1,45 +1,68 @@
...
 
+  PublicAccountsResult* = Result[seq[PublicAccount], string]
+
+  ChatAccountResult* = Result[accounts.Account, string]
+
 proc storeDerivedAccounts(self: StatusObject, id: UUID, keyUid: string,
   paths: seq[KeyPath], password, dir: string,
-  accountType: AccountType): PublicAccountResult {.raises: [Defect,
-  ref IOError, ref KeyError, ref OSError, SqliteError, ValueError].} =
+  accountType: AccountType): PublicAccountResult {.raises: [Defect].} =
```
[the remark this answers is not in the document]
emizzle: Correct, it is not needed. I've removed from here and another instance as well. Likely this was leftover from the first step changes being in place with all possible exceptions listed, then slowly chipping away each except from the list until only defect was left, at which point I should have just removed the pragma altogether.

status/api/accounts.nim
```diff
@@ -1,45 +1,68 @@
...
+  # ref IOError, ref KeyError, ref OSError, ValueError].} =
```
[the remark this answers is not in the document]
emizzle: Done.

status/api/accounts.nim
```diff
@@ -1,45 +1,68 @@
...
+  except PublicAccountDbError as e:
+    return PublicAccountResult.err "Failed to store derived accounts: " & e.msg
 
   let
-    defaultWalletAccountDerived = accountInfos[PATH_DEFAULT_WALLET]
+    defaultWalletAccountDerived = try: accountInfos[PATH_DEFAULT_WALLET]
+                                  except KeyError as e:
+                                    return PublicAccountResult.err "Error " &
+                                      "getting default wallet from derived " &
+                                      "accounts: " & e.msg
```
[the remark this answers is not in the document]
emizzle: Agreed, updated.

status/api/wallet.nim
emizzle: It means that there is no chance for that proc to raise a `Defect` based on the compiler's analysis. Because `{.push raises: [Defect].}` effectively adds a `{.raises: [Defect].}` to each proc/func/method/template/macro in the module, if there is a proc/func/method/template/macro that does not raise `Defect`, the `raises` pragma needs to be explicitly overridden for the compiler to understand.

status/private/chats.nim
emizzle: Good catch, updated.

status/private/identicon.nim
emizzle: Updated.

status/private/migration.nim
emizzle: Great catch, updated.

status/private/tx_history.nim
emizzle: Definitely agreed. There are no code paths from our public APIs to this module, nor any calls internally to this code. As discussed, we should repurpose it to be consumed by the client in a manner suited to the architecture of the client. In the case of the example chat client, we could start a long-running task that does this work.

emizzle:
> Note my comment above re: adding a WIP commit; should be squashed into the main commit before merge.

All squashed. Should be good to merge if approved.

status/api/wallet.nim
arnetheduck: > It means that there is no chance for that proc to raise a Defect based on the compiler's analysis. 

This unfortunately is not quite so - in nim 1.4, `raises: []` means that `Defect` can still be raised, while in nim 1.2 this transition was done only partially and some `Defect` trigger a compile-time error and some don't - we use `push raises: [Defect]` to prepare the code for nim 1.4 where it'll be replaced with `push raises: []` throughout, because this is the semantic that most closely matches where nim is heading.

emizzle: Makes sense Jacek, thanks so much for the info.

michaelsbradleyjr (approved):

PR #268, opened 2021-08-12
feat: add /sendtransaction command to TUI

richard-ramos:
Example command:
`/trx 0x30575c21a1Ec966Cb06eC9aa3dc697c8dECd0e14 0x0000000000000000000000000000000000000000 0 1 255 10000 0x7f7465737432000000000000000000000000000000000000000000000000000000600057 0 1`

Response from infura:
image

Receiving this message means that Infura understood the raw transaction that was signed and was able to get the `from` address from the signed transaction.

### DIFF
--- a/examples/client/tui/commands.nim
+++ b/examples/client/tui/commands.nim
@@ -4,6 +4,10 @@ import # std libs
 import # client modules
   ./common, ./macros, ./screen, ./tasks
 
+import # vendor libs
+  eth/common as eth_common, stew/byteutils
+
+
 export common, screen, strutils, tasks
 
 logScope:
@@ -67,7 +71,7 @@ proc command*(self: Tui, command: AddCustomToken) {.async, gcsafe,
 
   var parsedAddr: Address
   try:
-    parsedAddr = command.address.parseAddress
+    parsedAddr = common.parseAddress(command.address)
   except:
     self.wprintFormatError(getTime().toUnix,
       "Could not parse address, please provide an address in proper format.")
@@ -459,7 +463,7 @@ proc split*(T: type GetAssets, argsRaw: string): seq[string] =
 proc command*(self: Tui, command: GetAssets) {.async, gcsafe, nimcall.} =
   var parsedOwner: Address
   try:
-    parsedOwner = command.owner.parseAddress
+    parsedOwner = common.parseAddress(command.owner)
   except:
     self.wprintFormatError(getTime().toUnix,
       "Could not parse address, please provide an address in proper format.")
@@ -761,13 +765,7 @@ proc help*(T: type CallRpc): HelpText =
 
 proc new*(T: type CallRpc, args: varargs[string]): T =
   var rpcMethod = if args.len > 0: args[0] else: ""
-  var params = newJArray()
-  if args.len > 1:
-    try:
-      params = args[1..^1].join(" ").parseJson()
-    except:
-      params = newJNull()
-
+  var params = if args.len > 1: args[1..^1].join(" ") else: ""
   T(rpcMethod: rpcMethod, params: params)
 
 proc split*(T: type CallRpc, argsRaw: string): seq[string] =
@@ -779,11 +777,144 @@ proc command*(self: Tui, command: CallRpc) {.async, gcsafe,
     if command.rpcMethod == "":
       self.wprintFormatError(getTime().toUnix,
         "method cannot be blank, please provide an input.")
-    elif command.params.kind == JNull:
+      return
+
+    var params = newJArray()
+    if command.params != "":
+      try:
+        params = command.params.parseJson()
+      except:
+        self.wprintFormatError(getTime().toUnix,
+          "params must be a valid JSON")
+        return
+
+    asyncSpawn self.client.callRpc(command.rpcMethod, params)
+  except:
+    self.wprintFormatError(getTime().toUnix, "invalid arguments.")
+
+# SendTransaction --------------------------------------------------------------
+
+proc help*(T: type SendTransaction): HelpText =
+  let command = "sendtransaction"
+  HelpText(command: command, aliases: aliased.getOrDefault(command), parameters: @[
+    CommandParameter(name: "from", description: "The address for the sending account"),
+    CommandParameter(name: "to", description: "The destination address of the message, left empty for a contract-creation transaction."),
+    CommandParameter(name: "value", description: "The value transferred for the transaction in wei"),
+    CommandParameter(name: "maxPriorityFee", description: "Max priority fee in wei"),
+    CommandParameter(name: "maxFee", description: "Max fee in wei"),
+    CommandParameter(name: "gasLimit", description: "The amount of gas to use for the transaction (unused gas is refunded)."),
+    CommandParameter(name: "data", description: "ABI byte string containing the data of the function call on a contract, or contract-creation initialisation code."),
+    CommandParameter(name: "nonce", description: "Integer of the nonce"),
+    CommandParameter(name: "password", description: "Account password")
+    ], description: "Sends an EIP1559 transaction")
+
+proc new*(T: type SendTransaction, args: varargs[string]): T =
+  T(fromAddress: args[0],
+    toAddress: args[1],
+    value: args[2],
+    maxPriorityFee: args[3],
+    maxFee: args[4],
+    gasLimit: args[5],
+    payload: args[6],
+    nonce: args[7],
+    password: args[8])
+
+proc split*(T: type SendTransaction, argsRaw: string): seq[string] =
+  let args = argsRaw.split(" ")
+  var values: array[9, string]
+  for idx, val in args[0..min(high(values), high(args))].pairs:
+    values[idx] = val
+  @values
+
+proc command*(self: Tui, cmd: SendTransaction) {.async, gcsafe,
+  nimcall.} =
+  try:
+    var fromAddress: EthAddress
+    try:
+      fromAddress = eth_common.parseAddress(cmd.fromAddress)
+    except:
       self.wprintFormatError(getTime().toUnix,
-        "params cannot be blank and must be a valid JSON, please provide an input.")
+        "Could not parse `from` address, please provide an address in proper format.")
+      return
+
+    var toAddress: Option[EthAddress]
+    if cmd.toAddress != "":
+      try:
+        toAddress = some(eth_common.parseAddress(cmd.toAddress))
+      except:
+        self.wprintFormatError(getTime().toUnix,
+          "Could not parse `to` address, please provide an address in proper format.")
+        return
     else:
-      asyncSpawn self.client.callRpc(command.rpcMethod, command.params)
+      toAddress = none(EthAddress)
+
+    var value = 0.u256
+    if cmd.value != "":
+      try:
+        value = cmd.value.u256
+      except:
+        self.wprintFormatError(getTime().toUnix,
+        "Could not parse `value`, please provide a valid numeric value (wei).")
+        return
+
+    var gasLimit = int64(0)
+    if cmd.gasLimit != "":
+      try:
+        gasLimit = cmd.gasLimit.parseBiggestInt
+      except:
+        self.wprintFormatError(getTime().toUnix,
+        "Could not parse `gasLimit`, please provide a valid numeric value.")
+        return
+
+    var maxPriorityFee = int64(0)
+    if cmd.maxPriorityFee != "":
+      try:
+        maxPriorityFee = cmd.maxPriorityFee.parseBiggestInt
+      except:
+        self.wprintFormatError(getTime().toUnix,
+        "Could not parse `maxPriorityFee`, please provide a valid numeric value.")
+        return
+
+    var maxFee = int64(0)
+    if cmd.maxFee != "":
+      try:
+        maxFee = cmd.maxFee.parseBiggestInt
+      except:
+        self.wprintFormatError(getTime().toUnix,
+        "Could not parse `maxFee`, please provide a valid numeric value.")
+        return
+
+    var payload:seq[byte] = @[]
+    if cmd.payload != "":
+      try:
+        payload = cmd.payload.hexToSeqByte
+      except:
+        self.wprintFormatError(getTime().toUnix,
+        "Could not parse `payload`, please provide a valid hex string.")
+        return
+
+    var nonce = uint64(0)
+    if cmd.nonce != "":
+      try:
+        nonce = uint64(cmd.nonce.parseBiggestInt)
+      except:
+        self.wprintFormatError(getTime().toUnix,
+        "Could not parse `nonce`, please provide a valid numeric value.")
+        return
+
+    asyncSpawn self.client.sendTransaction(
+      fromAddress,
+      Transaction(
+        txType: TxType.TxEip1559,
+        to: toAddress,
+        value: value,
+        maxPriorityFee: maxPriorityFee,
+        maxFee: maxFee,
+        gasLimit: gasLimit,
+        payload: payload,
+        nonce: nonce,
+      ),
+      cmd.password)
   except:
     self.wprintFormatError(getTime().toUnix, "invalid arguments.")
 
--- a/status/api/provider.nim
+++ b/status/api/provider.nim
@@ -1,17 +1,22 @@
 {.push raises: [Defect].}
 
 import # std libs
-  std/json
+  std/[json, sequtils, sugar]
 
 import # vendor libs
-  chronos
+  chronicles, chronos,
+  eth/common as eth_common,
+  eth/[common/transaction, keys, rlp],
+  stew/byteutils
 
 import # status modules
-  ../private/callrpc,
+  ../private/[accounts/accounts, accounts/generator/generator, callrpc, settings],
   ./common
 
 type
   CallRpcResult* = Result[JsonNode, string]
+
+  SendTransactionResult* = Result[JsonNode, string]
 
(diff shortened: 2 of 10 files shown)


examples/client/tui/commands.nim
```diff
@@ -788,6 +792,48 @@ proc command*(self: Tui, command: CallRpc) {.async, gcsafe,
...
+  let fromAddress = eth_common.parseAddress(args[0])
+  let transaction = Transaction(
+    txType: TxType.TxEip1559,
+    to: some(eth_common.parseAddress(args[1])),
+    value: args[2].u256,
+    maxPriorityFee: args[3].parseBiggestInt,
+    maxFee: args[4].parseBiggestInt,
+    gasLimit: args[5].parseBiggestInt,
+    payload: args[6].hexToSeqByte,
+    nonce: uint64(args[7].parseBiggestInt)
```
richard-ramos, line 821: @emizzle, @michaelsbradleyjr  what's the best way to validate that the arguments have the correct format? Here I receive a combination of eth address, integers, and hex strings

michaelsbradleyjr: Good question, made me realize something that I missed when reviewing your PR for the `/call` command:

https://github.com/status-im/nim-status/blob/feat/send-transaction/examples/client/tui/common.nim#L44-L45
> all fields on types that derive from Command should be of type `string`

The way the command parser is currently implemented, neither a command's `split` nor `new` procs are in a good position to do validation and if they raise the program will crash. See this comment at the top of the commands module:

https://github.com/status-im/nim-status/blob/feat/send-transaction/examples/client/tui/commands.nim#L18-L30

Commonly, that means letting a field for a missing parameter be an empty string, and with all fields being strings and minimal logic in those parts, the command parser is unlikely to crash.

The best place to do validation and convert from string to non-string is therefore in the `command` proc, and looking at what you have so far, it seems you're going that direction.

I suggest creating a:
```nim
proc validate(command: SendTransaction): SendTransactionResult
```

And then in the command proc you can do something like:
```nim
let cmdResult = SendTransaction.validate()

if cmdResult.isErr:
  self.wprintFormatError(getTime().toUnix, "Invalid arguments: " & cmdResult.error)
else:
  let cmd = cmdResult.get
  asyncSpawn self.client.sendTransaction(
    eth_common.parseAddress(cmd.fromAddress),
    Transaction(
      txType: TxType.TxEip1559,
      to: some(eth_common.parseAddress(cmd.to)),
      value: cmd.value.u256,
      maxPriorityFee: cmd.maxPriorityFee.parseBiggestInt,
      maxFee: cmd.maxFee.parseBiggestInt,
      gasLimit: cmd.gasLimit.parseBiggestInt,
      payload: cmd.payload.hexToSeqByte,
      nonce: uint64(cmd.nonce.parseBiggestInt),
    ),
    cmd.password
)
```

I will understand if you find it a bit awkward. I think the way the command parser works can be improved in a future PR/s. For example,  in addition to `help`, `new`, `split`, and `command` procs for each command type, there could be a `validate` command and it would automatically get used by the [parser](https://github.com/status-im/nim-status/blob/feat/send-transaction/examples/client/tui/parser.nim). If `validate` results in an `isErr` then the parser would take care of printing the error, instead of that logic having to be implemented by hand in each `command` proc.

That still leave the awkwardness re: converting from string to non-string, and often times there is overlap between validation logic and conversion logic. I need to think more about the relationship between the parser and dispatch and how we can make that part as pain-free as possible.

michaelsbradleyjr: Hmm, I guess for now, the `validate` proc for `SendTransaction` doesn't have to return a `SendTransaction` instance wrapped in `Result`. The instance wrapped in `Result` can instead be of the type that has the fields converted as you want.

So the last part would be:
```nim
asyncSpawn self.client.sendTransaction(cmd.fromAddress, cmd.transaction, cmd.password)
```

michaelsbradleyjr: That approach will work generally, but wouldn't lend itself to being plugged into the parser because of how the types have to match up for dispatch.

I'll think about this some more, there's probably a fun and clever way to make it pretty clean with the help of one/more macros and maybe some `cast`ing.

richard-ramos: Alright, I did the change although I did the validation in place instead of having a separate proc (to not deviate much from what is done in the other commands). In this PR I also changed the CallRpc command so the params are a string instead of a Json

status/private/callrpc.nim
```diff
@@ -85,13 +85,14 @@ proc newWeb3*(settings: Settings): Web3 {.raises: [CatchableError, Defect,
 
 proc callRpc*(web3Conn: Web3, rpcMethod: RemoteMethod, params: JsonNode):
   Future[Response] {.async, raises: [Defect, CatchableError].} =
-  const errorMsg = "Error calling RPC method"
   try:
     return await web3Conn.provider.call($rpcMethod, params)
   except CatchableError as e:
-    raise (ref Web3Error)(parent: e, msg: errorMsg)
+    let err = getCurrentExceptionMsg()
+    raise (ref Web3Error)(parent: e, msg: getCurrentExceptionMsg())
```
michaelsbradleyjr, line 92: Is there a problem with doing `msg: err` since in the line above you `let err = getCurrentExceptionMsg()`?

Also, I'm curious why it's important in this case (and below) to do `getCurrentExceptionMsg()` instead of referring to `e.msg`?

status/api/provider.nim
```diff
@@ -23,3 +28,59 @@ proc callRpc*(self: StatusObject, rpcMethod: string, params: JsonNode):
...
+  if not self.isLoggedIn:
+    return SendTransactionResult.err "Not logged in. Must be logged in"
+
+  try:
+    let accounts = self.userDb.getWalletAccounts()
+    let account = accounts.filter(x => x.address == Address(fromAddress))
+    if account.len == 0:
+      return SendTransactionResult.err "Wallet address not found"
+
+    let loadAccountResult = self.accountsGenerator.loadAccount(account[0].address, password,dir)
```
michaelsbradleyjr, line 58: It's probably a matter of my being ignorant re: how some of the crypto parts work, but is loading the keystore from disk every time simply how we have to do it? Does status-go do the same?

Or is it a matter of something/s re: nim-status' sqlcipher database and related procs not being wired up in a way that would allow us to use info from the open database instead of loading the keystore?

status/private/callrpc.nim
```diff
@@ -85,13 +85,14 @@ proc newWeb3*(settings: Settings): Web3 {.raises: [CatchableError, Defect,
...
   except ValueError as e:
-    raise (ref Web3Error)(parent: e, msg: errorMsg)
+    let err = getCurrentExceptionMsg()
+    raise (ref Web3Error)(parent: e, msg: getCurrentExceptionMsg())
```
michaelsbradleyjr, line 95: Same questions apply here.

status/private/callrpc.nim
```diff
@@ -105,10 +106,11 @@ proc callRpc*(web3Conn: Web3, rpcMethod: string, params: JsonNode):
   except:
     return %* {"code": -32601, "message": "the method " & rpcMethod & " does not exist/is not available"}
 
-  const errorMsg = "Error calling RPC method"
   try:
     return await web3Conn.provider.call(rpcMethod, params)
   except CatchableError as e:
-    raise (ref Web3Error)(parent: e, msg: errorMsg)
+    let err = getCurrentExceptionMsg()
+    raise (ref Web3Error)(parent: e, msg: getCurrentExceptionMsg())
```
michaelsbradleyjr, line 113: And here.

status/private/callrpc.nim
```diff
@@ -105,10 +106,11 @@ proc callRpc*(web3Conn: Web3, rpcMethod: string, params: JsonNode):
...
   except ValueError as e:
-    raise (ref Web3Error)(parent: e, msg: errorMsg)
+    let err = getCurrentExceptionMsg()
+    raise (ref Web3Error)(parent: e, msg: getCurrentExceptionMsg())
```
michaelsbradleyjr, line 116: And here.

examples/client/tui/commands.nim
```diff
@@ -780,11 +778,151 @@ proc command*(self: Tui, command: CallRpc) {.async, gcsafe,
...
+    maxPriorityFee: args[3],
+    maxFee: args[4],
+    gasLimit: args[5],
+    payload: args[6],
+    nonce: args[7],
+    password: args[8])
+
+proc split*(T: type SendTransaction, argsRaw: string): seq[string] =
+  let args = argsRaw.split(" ")
+  let
```
michaelsbradleyjr, line 825: What if instead of all the `if/else`, an array.of string with length 9 is initialized? Then you could do `for a in args: ...` and overwrite the empty strings, and return the array converted to a seq.

status/private/callrpc.nim
richard-ramos: It's a mistake. I'll fix it

status/api/provider.nim
richard-ramos: I was looking at the database and it seems that the private keys are not stored. This is the reason why I'm loading the keystore from disk instead of obtaining the info from the DB.

examples/client/tui/commands.nim
```diff
@@ -780,11 +778,144 @@ proc command*(self: Tui, command: CallRpc) {.async, gcsafe,
...
+    payload: args[6],
+    nonce: args[7],
+    password: args[8])
+
+proc split*(T: type SendTransaction, argsRaw: string): seq[string] =
+  let args = argsRaw.split(" ")
+  var values: array[9, string]
+  for n in 0 .. min(args.len - 1, values.len):
+    values[n] = args[n]
+  @values
```
michaelsbradleyjr, line 828:
```suggestion
  for idx, val in args.pairs:
    values[idx] = val
  @values
```

`pairs` is a nice way to get an index paired with a value for iterating over a seq/array.

I tested locally that the suggested change achieves the desired result.

michaelsbradleyjr (changes requested):

examples/client/tui/commands.nim
michaelsbradleyjr: I realize there was a bounds-checking flaw in my original suggested changes, fixed it.

status/api/provider.nim
```diff
@@ -23,3 +28,59 @@ proc callRpc*(self: StatusObject, rpcMethod: string, params: JsonNode):
...
+  var r = toRaw(s)
+  let v = r[64]
+
+  tr.R = fromBytesBE(UInt256, r.toOpenArray(0, 31))
+  tr.S = fromBytesBE(UInt256, r.toOpenArray(32, 63))
+
+  tr.V = int64(v)
+
+  if tr.txType == TxType.TxLegacy:
+    tr.V += 27 # TODO! Complete this
```
emizzle, line 45: I wonder if we should move this to `status/private/callrpc.nim`? Maybe `callrpc.nim` isn't the best name 🤔

status/api/provider.nim
```diff
@@ -23,3 +28,59 @@ proc callRpc*(self: StatusObject, rpcMethod: string, params: JsonNode):
...
+      txType: transaction.txType,
+      to: transaction.to,
+      value: transaction.value,
+      maxPriorityFee: transaction.maxPriorityFee,
+      maxFee: transaction.maxFee,
+      gasLimit: transaction.gasLimit,
+      payload: transaction.payload,
+      nonce: transaction.nonce,
+      chainId: ChainId(network.get.config.networkId)
+    )
```
emizzle, line 80: Why the need to rebuild the `Transaction` here? Could we instead accept a mutable `Transaction` object in the signature?

emizzle (approved):
Great job as always Richard! Nothing that would stop an approval, but a question and a suggestion.

status/api/provider.nim
richard-ramos: I tried using `var` but the compiler complained: `Error: invalid type: 'var Transaction' in this context: 'ref SendTransactionTaskArg:ObjectType' for var` . Not sure how to fix it

michaelsbradleyjr (approved):

status/api/provider.nim
emizzle: @michaelsbradleyjr, I believe this has something to do with either the task macro not allowing `var` arg types or `json_serialization` not being able to decode it properly (or a combination of both). What do you think?

emizzle: @richard-ramos, easiest to simply create a mutable copy like so. Can also remove the settings/network retrieval above it.

```suggestion
    var trx = transaction
```

michaelsbradleyjr: Nice catch, @emizzle. Afaict, the `type Transaction` we're working with is defined in `vendor/nim-waku/vendor/nim-eth/eth/common/eth_types.nim` using `Transaction* = object`.

So on assignment to `var trx` it gets copied and in the context of the body of `proc sendTransaction` that's enough to satisfy the compiler.

status/api/provider.nim
michaelsbradleyjr: Perhaps `status/api/provider.nim` and `status/private/callrpc.nim` could be renamed `status/api/ethereum_rpc.nim` and `status/private/ethereum_rpc.nim`, respectively?

I don't think it's critical for this PR, but definitely something good to think about.

If we wanted to be really forward-looking, we could have:
* `status/api/ethereum.nim`
* `status/api/ethereum/rpc.nim`
* `status/private/ethereum.nim`
* `status/private/ethereum/rpc.nim`

What I mean is that nim-status might embed [fluffy](https://our.status.im/nimbus-fluffly/) at some point, in which case the library's users could choose to `import status/api/ethereum/fluffy` and use direct calls, or instead choose to `import status/api/ethereum/rpc` and make calls to some endpoint, e.g. an Infura endpoint or a geth node they're running on their home network.

status/api/provider.nim
emizzle: Do you know why we can't pass in a mutable `Transaction` object in to the proc?

michaelsbradleyjr: I don't know, not exactly. I'll need to do some experiments tomorrow.

examples/client/tui/commands.nim
```diff
@@ -780,11 +778,144 @@ proc command*(self: Tui, command: CallRpc) {.async, gcsafe,
...
+
+    asyncSpawn self.client.callRpc(command.rpcMethod, params)
+  except:
+    self.wprintFormatError(getTime().toUnix, "invalid arguments.")
+
+# SendTransaction --------------------------------------------------------------
+
+proc help*(T: type SendTransaction): HelpText =
+  let command = "sendtransaction"
+  HelpText(command: command, aliases: aliased[command], parameters: @[
```
michaelsbradleyjr, line 800:
```suggestion
  HelpText(command: command, aliases: aliased.getOrDefault(command), parameters: @[
```

My apologies for another requested change; this one is re: bugfix #277, which just landed.

michaelsbradleyjr (changes requested):

status/api/provider.nim
```diff
@@ -23,3 +28,59 @@ proc callRpc*(self: StatusObject, rpcMethod: string, params: JsonNode):
...
+proc sendTransaction*(self: StatusObject, fromAddress: EthAddress, transaction: Transaction, password: string, dir: string):
+  Future[SendTransactionResult] {.async.} =
+  if not self.isLoggedIn:
+    return SendTransactionResult.err "Not logged in. Must be logged in"
+
+  try:
+    let accounts = self.userDb.getWalletAccounts()
+    let account = accounts.filter(x => x.address == Address(fromAddress))
+    if account.len == 0:
+      return SendTransactionResult.err "Wallet address not found"
```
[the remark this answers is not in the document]
michaelsbradleyjr: It was discovered today that when using `getWalletAccount` in `status/private/accounts/accounts.nim` (as this PR does presently), then because of the [`AND wallet = 0`](https://github.com/status-im/nim-status/blob/master/status/private/accounts/accounts.nim#L112) condition, it's not possible get the account that corresponds to the `Status account`.

Here's how to replicate:
* delete the client's data directory, if it exists
* build and run the client
* do `/create 1` and `/login 1 1`
* do `/wallets` and copy/paste address for `Status account` into:
```
/trx [address] 0x0000000000000000000000000000000000000000 0 1 255 10000 0x7f7465737432000000000000000000000000000000000000000000000000000000600057 0 1
```

Then run that command. In the TUI an error will be displayed: `Error Wallet address not found`.

Next do `/addwallet 1`, get the address for the new wallet from `/wallets` and run the `/trx ...` command with that address. An error will be printed; you'll see something like:
```
Error {"code":-32602,"message":"transaction could not be decoded: chainId is required for EIP-2718 style signatures"}
```

Which means the command and nim-status lib did work as expected.

Maybe `getWalletAccount` should be checking for `AND chat = 0`?

michaelsbradleyjr:
I rebased this PR branch against `master`, made some changes suggested above and added commits. When the lights go green in CI, I intend to squash the commits into one.

However, I'll hold off on merging it pending discussion/decision re: https://github.com/status-im/nim-status/pull/268#discussion_r692363992.

michaelsbradleyjr (approved):

status/api/provider.nim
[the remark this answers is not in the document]
michaelsbradleyjr: status-go checks `chat = 0` in its `GetWalletAddresses` function:

https://github.com/status-im/status-go/blob/develop/multiaccounts/accounts/database.go#L634-L635

So we can/should do likewise. I'm going to merge this PR and create another re: the fix for `getWalletAccount`.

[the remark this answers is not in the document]
michaelsbradleyjr: See #288.

PR #277, opened 2021-08-17
fix: help command should not crash if some commands don't have alias

### LINKED ISSUE #276: help command is crashing the process if one or more commands don't have alias/es

michaelsbradleyjr:
Also fix `proc help` for `AddWalletWatchOnly`, which was incorrectly using `"addwalletseed"` instead of `"addwalletwatch"`.

Closes #276.


emizzle (approved):
Nice!

Issue #278, opened 2021-08-18
Support switching to testnet


emizzle:
Should support `/listnetworks` to list available networks.

Should support something like`/switchnetwork <index>`, which will set the current network to the network config with that index.

Might be a good idea to keep a list of network in a database table instead of just in settings. Could have a relationship between contracts, networks, and contracts_networks mapping table. Needs investigation.

emizzle:
It's far easier to develop stickers support on testnet as we don't need real eth to execute txs.

PR #297, opened 2021-08-25
feat: allow network to be switched

emizzle:
Closes: #278.

Add `/switchnetwork` command that allows a user to switch their upstream RPC network.

feat: add /listnetworks command (aliased to `/networks`)
Lists all upstream networks from settings.

Also included in this PR is some alphabetical reorganisation.

NOTE: this PR could be squashed before merge.

### DIFF
--- a/examples/client/client.nim
+++ b/examples/client/client.nim
@@ -114,11 +114,11 @@ proc getAssets*(self: Client, owner: Address) {.async.} =
 proc getCustomTokens*(self: Client) {.async.} =
   asyncSpawn getCustomTokens(self.taskRunner, status)
 
-proc getTopics(self: Client): Future[seq[ContentTopic]] {.async, gcsafe.} =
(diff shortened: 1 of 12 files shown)


michaelsbradleyjr:
I've rebased this PR's branch against the branch for PR #271 and changed the base branch of this PR to #271's branch.

Something quite strange happened during the rebase, never experienced it previously. My editor started warning me about changes on disk that might need to be reloaded (but it's always handled that automatically in the past). I made what I thought were the correct choices but the end result on my first round of rebasing had some seriously fubar stuff in client/tasks, tui/actions, and tui/commands.

I fixed it all — the tests are passing and the example client builds and runs correctly. I visually scanned over every line of the modules mentioned above and didn't spot any lingering rebase mistakes. I think *maybe* some line-length cleanup changes in tui/commands got lost along the way, but the clean alphabetization is still intact.

PR #303, opened 2021-09-13
refactor: simplify database migrations

### LINKED ISSUE #206: Create user db schema for use with db initialization
Create a fresh status-go db, and export the db without the accounts (can be done with DB Browser for SQLite. It should have default settings populated.

Migrations will then only be needed for incremental changes after nim-status is being actively used.

### LINKED ISSUE #208: Create accounts db schema for use with db initialization

michaelsbradleyjr:
Prior to this PR, nim-status relied on a database migrations setup that involved `.nim` codegen in a `make` target; the SQL migration scripts involved in that target were copies from the [status-go](https://github.com/status-im/status-go) repo.

nim-status had not kept up with migration scripts added to status-go; also, the database migrations logic of status-go had changed (relative to nim-status' logic) such that simply copying more recent migrations scripts into nim-status would not result in successful database setup.

The changes introduced in this PR represent a different approach:

Schemas for fresh "accounts" and "user" databases created by latest status-go were exported and copied into nim-status (see: `status/private/database/accounts/000000001.sql` and `status/private/database/user/000000001.sql` in this PR's branch), with the intention of using them as a new starting point for nim-status' database setup. In the future, nim-status may diverge from status-go at the database level (removing tables, columns, etc., while introducing new ones). In other words, database-level compatibility with status-go is not a goal, though protocol-level compatibility remains a goal (it's a *must*, of course).

The new migrations approach involves use of SQLite's [user_version](https://www.sqlite.org/pragma.html#pragma_user_version) pragma and hashing of database schemas so that it won't be possible to inadvertently run an incompatible migrations script.

We should keep in mind the notes in status-go's [migrations/sqlite/README.md](https://github.com/status-im/status-go/blob/develop/protocol/migrations/sqlite/README.md) and [migrations/sqlite/doc.go](https://github.com/status-im/status-go/blob/develop/protocol/migrations/sqlite/doc.go) while revising our `status/private/database/user/000000001.sql`. It may be that we'll want/need to introduce separete databases or namespacing within the user database.

Closes #206.
Closes #208.


status/private/database.nim
```diff
@@ -1,69 +1,169 @@
...
-      let definition = acc_migration.newMigrationDefinition()
-      discard ?dbConn.migrate(definition).mapErrTo(DbError.MigrationError)
-    ok dbConn
+    paths = sorted(paths, system.cmp[string])
+
+    for path in paths:
+      result.add path.staticRead.strip.replace("\r\n", "\n").replace("\r", "\n")
+
+  except Exception:
+    raise newException(Defect, "unable to read .sql files in " & dir)
```
emizzle, line 32: We have a `StatusDefect` in `./common`, maybe that would be a better choice here?

Was `Exception` was used on purpose to catch both `CatchableError` and `Defect`? If so, we should probably simply catch those two exception types, ie:
```nim
except CatchableError, Defect:
  # ...
```

status/private/database.nim
```diff
@@ -1,69 +1,169 @@
...
 
-proc hash(password: string): string {.raises: [].} =
-  "0x" & toUpper($keccak_256.digest(password))
+when accountsDbSchemaHashes.len != accountsDbMigrations.len:
+  {.fatal:
+    "accountsDbSchemaHashes.len must be equal to accountsDbMigrations.len!".}
 
-proc initDb*(path, password: string): DbResult[DbConn] =
+when userDbSchemaHashes.len != userDbMigrations.len:
+  {.fatal: "userDbSchemaHashes.len must be equal to userDbMigrations.len!".}
```
emizzle: This is really cool!

emizzle (approved):
Really well done, Micahel! I like how the schema is hashed on the fly to compare to the migration to be run. I need to re-review this with fresh eyes in the morning just so I have a better understanding of the mechanics, but all looks good to merge from the first review.

status/private/database.nim
michaelsbradleyjr: The compiler was complaining about `Exception` being unlisted in `raises`. Note that the context is `staticReadSqlFiles`, which is compile-time only, so I was thinking the most important thing in that situation is to ensure compilation fails (by raising a `Defect`) and including a relevant error message.

